PR #2, opened 2020-02-03
def validate = Try(transform)

htmldoug:
@russellremple @jducoeur


SCKelemen (approved):

russellremple (approved):
lgtm. Then later if we have a validate that returns all errors, we could call it "validateFully" or something like that.

PR #4, opened 2020-02-04
differences.md fixes

danny-goldberg-rally:


differences.md
```diff
@@ -18,11 +18,11 @@ play-json always writes a `"key":"value"` pair regardless of whether it's the ca
 
 weepickle works [differently](http://www.lihaoyi.com/upickle/#Defaults). If a field is missing upon deserialization, weepickle uses the default value if one exists. If a field at serialization time has the same value as the default, weepickle leaves it out of the serialized blob.
 
-OpenAPI 3 [has support](https://swagger.io/specification/#schemaObject) for weepickle's behavior:
```
danny-goldberg-rally, line 21: this made it sound like play-json doesn't have support for weepickle's default behavior

htmldoug: It doesn't, does it? You'd have to drop down to custom `Writes` to achieve what weePickle can do with the macros alone.

htmldoug (approved):
thanks!

differences.md
danny-goldberg-rally: play-json does support `weePickle`'s default behavior, but doesn't support `uPickle`'s default behavior

russellremple (approved):
lgtm

PR #5, opened 2020-02-04
When decoding for a byte array, treat strings as b64-encoded bytes

russellremple:
@htmldoug @jducoeur

### DIFF
--- a/implicits/src/com/rallyhealth/weepickle/v1/implicits/DefaultTos.scala
+++ b/implicits/src/com/rallyhealth/weepickle/v1/implicits/DefaultTos.scala
@@ -2,7 +2,7 @@ package com.rallyhealth.weepickle.v1.implicits
 
 import java.net.URI
 import java.time.Instant
-import java.util.UUID
+import java.util.{Base64, UUID}
 import java.util.concurrent.TimeUnit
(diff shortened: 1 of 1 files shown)


htmldoug (approved):

implicits/src/com/rallyhealth/weepickle/v1/implicits/DefaultTos.scala
```diff
@@ -256,6 +256,9 @@ trait DefaultTos extends com.rallyhealth.weepickle.v1.core.Types with Generated
 
         def subVisitor = implicitly[To[T]]
       }
+      override def visitString(cs: CharSequence): Array[T] = {
+        Base64.getDecoder.decode(cs.toString).asInstanceOf[Array[T]]
```
danny-goldberg-rally, line 260: why?

russellremple: Doug assumed binary data serialized to Json should be encoded as b64 strings. So if you serialize binary data to Json and then attempt to deserialize back to the source structure, the reverse assumption needs to be made or deserialization fails.

PR #16, opened 2020-02-26
MapKeys (snake_case) and wrap a Visitor to return all errors

russellremple:
Maybe no one will ever need this, but just in case they do, here it is. The earlier version was on the "From" side and was specific to Json Value. This one is on the "To" side, so it is more flexible -- you can wrap any Visitor and get all the errors. Moving it here in weePickle from lib-weepickle-extras because there are no Rally-specific components. If you don't want it here, we can make it a "spice". Tests are in the weejson play project because it was easier to satisfy dependencies that way. 

@htmldoug @vito-c @jducoeur


htmldoug (commented):
Neat. A few questions.
1. Possible to avoid duplication in the copy pasted sections?
2. Are there any cases where a stateful visitor might really want to abort but wouldn't be able to? For example, I'm thinking about the ToJson that auto closes after a single value. Is it desirable that all future method calls would throw and add a path? I'm trying to think of less corner casey situations. Feels like there should be some but none immedialy come to mind.

russellremple:
> 1. Possible to avoid duplication in the copy pasted sections?

I thought about changing Visitor.MapTo to have a function from name rather than from value. That would allow the implementation to catch any thrown exception. (I.e., change `mapFunction(v: V): Z` to `mapFunction(v: => V): Z`, but leaving `def mapNonNullsFunction(t: V): Z` alone.) That way the client could override mapFunction if they wanted to catch exceptions around building V, or just override mapNonNullsFunction if they didn't. But thought that might be too much of a divergence from how it is used everywhere. And maybe there'd be a tiny performance overhead of function boxing for call by name where it isn't done today. 

Similarly the code for HasPath is lifted pretty directly from JsonPointerVisitor, but it has a pretty different implementation of `wrap`. I suppose the original trait could be changed to public and then extended in ValidatingVisitor with that method overridden (or maybe introduce a new common ancestor), but in the end you save only about 30 duplicated lines. 

So I didn't do either, and wound up with a kind of mash up that you see here: a little MapTo, a little JsonPointerVisitor, and a lot of exception catching!

> 2. Are there any cases where a stateful visitor might really want to abort but wouldn't be able to? For example, I'm thinking about the ToJson that auto closes after a single value. Is it desirable that all future method calls would throw and add a path? I'm trying to think of less corner casey situations. Feels like there should be some but none immediately come to mind.

If I understand correctly, we would have to differentiate which exceptions indicate the Visitor "wants to abort" and which do not? I suppose we could introduce a kind of poison pill exception type that, if thrown, would abort parsing rather than being trapped as an error. But I'm not sure that would come up in practice (especially if documents are relatively small). I'd say those folks should probably just keep using the "normal" Visitor behavior of throwing on the first error... so you either get the first error, or you get all the errors, but nothing in between.

core/src/com/rallyhealth/weepickle/v1/core/ValidatingVisitor.scala
```diff
@@ -0,0 +1,236 @@
...
+  * A slight variation of JsonPointerVisitor, where we
+  * add a *sequence* of JSON Pointers to the result of the delegate Visitor
+  * rather than throwing a single exception for the first error encountered by the delegate.
+  *
+  * JSON Pointer is standardized by RFC 6901 and commonly used by JSON Schema.
+  *
+  * Useful for debugging failures.
+  * Adds ~10% overhead depending on the parser. [?]
+  *
+  * @see https://tools.ietf.org/html/rfc6901
```
vito-c: TIL :)

vito-c, line 27: TIL :)

core/src/com/rallyhealth/weepickle/v1/core/ValidatingVisitor.scala
```diff
@@ -0,0 +1,236 @@
...
+
+/**
+  * A slight variation of JsonPointerVisitor, where we
+  * add a *sequence* of JSON Pointers to the result of the delegate Visitor
+  * rather than throwing a single exception for the first error encountered by the delegate.
+  *
+  * JSON Pointer is standardized by RFC 6901 and commonly used by JSON Schema.
+  *
+  * Useful for debugging failures.
+  * Adds ~10% overhead depending on the parser. [?]
```
htmldoug: How does this compare to JsonPointerVisitor?

I'm curious about the cost of the extra Either, Try, and ListBuffers.

russellremple: I did not benchmark. These comments were carried forward from JsonPointerVisitor.

russellremple: Here are some microbenchmarks against a 640 KB, 24 K line Json with two errors:

```[info] Benchmark               Mode  Cnt    Score   Error  Units
[info] BenchTest.play_All     thrpt   20   57.468 ± 1.367  ops/s
[info] BenchTest.wee_1_First  thrpt   20  229.404 ± 5.225  ops/s
[info] BenchTest.wee_2_All    thrpt   20  184.232 ± 3.377  ops/s
```
So seems that it adds about 20% overhead (about 1 ms)

htmldoug:
Sounds reasonable.

core/src/com/rallyhealth/weepickle/v1/core/ValidatingVisitor.scala
```diff
@@ -0,0 +1,243 @@
...
+
+    def returningAllErrors: Visitor[T, Either[Errors, J]] = new ValidatingVisitor(delegate, RootHasPath)
+  }
+
+  /**
+    * Internally, the paths form a linked list back to the root by the visitors themselves.
+    * Compared to something like a List[String] or List[Object], this does not require
+    * extra String allocation or boxing unless we actually ask for the path.
+    */
+  trait HasPath {
```
htmldoug: Can `HasPath` and `HasMultiplePaths` be private?

core/src/com/rallyhealth/weepickle/v1/core/Visitor.scala
```diff
@@ -284,6 +285,33 @@ object Visitor {
...
+  class MapKeys[T, J](delegatedTo: Visitor[T, J], keyMapper: CharSequence => CharSequence) extends Delegate[T, J](delegatedTo) {
+
+    override def visitObject(length: Int): ObjVisitor[T, J] =
+      new MapKeyObjDelegate[T, J](delegatedTo.visitObject(length), keyMapper)
+
+    override def visitArray(length: Int): ArrVisitor[T, J] =
+      new MapKeyArrDelegate[T, J](delegatedTo.visitArray(length), keyMapper)
+  }
+
+  class MapKeyObjDelegate[T, J](objVisitor: ObjVisitor[T, J], keyMapper: CharSequence => CharSequence) extends ObjDelegate[T, J](objVisitor) {
```
htmldoug: Can we make "Key" plural for consistency with MapKeys?

core/src/com/rallyhealth/weepickle/v1/core/Visitor.scala
```diff
@@ -102,6 +102,7 @@ trait Visitor[-T, +J] extends AutoCloseable {
     override def mapFunction(v: J): Z = f(v)
     def mapNonNullsFunction(v: J): Z = f(v)
   }
+  def mapKeys(f: CharSequence => CharSequence): Visitor[T, J] = new Visitor.MapKeys[T, J](Visitor.this, f)
```
htmldoug: I'd rather keep this method definition outside of the minimal Visitor interface. Fine as an implicit extension method though. Should probably include "recursive" in the name. I could see use cases either way.

core/src/com/rallyhealth/weepickle/v1/core/ValidatingVisitor.scala
russellremple: done

core/src/com/rallyhealth/weepickle/v1/core/Visitor.scala
russellremple: done

core/src/com/rallyhealth/weepickle/v1/core/Visitor.scala
russellremple: I created VisitorImplicits to house typeclass for mapKeys. I did not include "recursive" in the name because Delegate and MapTo (accessed by "map" function) are also recursive and don't have "recursive" in their names, and wanted to stay consistent with that.

perryjp:
what's the status on this?

htmldoug:
`mapKeys` is pretty specific to add to the core API. Can be recursive or not and there are use cases for both. MsgPack can even allow non-String keys, although I don't care a ton about supporting that. I'm okay with adding well-named decorator class(es) like `MapKeysRecursive`, and even extension methods, but not adding methods to the traits directly for something so specific.

For "return all errors", I'd like to see this incubated somewhere. The capability looks useful, but the pickle family was built around the idea of exceptions terminating the parsing. I'm not sure what the consequences are for continuing. If we find out later that it's a bad idea, I can't really remove it. weePickle's primary purpose is to offer a stable API. There's also no reason it has to be in core, although I'd be open to pulling it in after it's a bit more battle tested.

plokhotnyuk:
Please consider making the "returning of all errors" feature opted out by default. It is hard to make it efficient enough for all worst cases upfront, but for open systems, it can be yet another source of vulnerabilities under DoS/DoW attacks.

So it looks like a nice feature for more productive development and maintenance [when errors are exceptional](https://shipilev.net/blog/2014/exceptional-performance/). But when carefully selected malicious messages hit your service that opened to the wild wide world it can be sensitive for the business.

PR #17, opened 2020-03-02
Add ZonedDateTime to implicits and readme

robert-j-miller:
We support Date, Instant, LocalDate, LocalTime, LocalDateTime, and OffsetDateTime already. This adds ZonedDateTime to the list of types supported implicitly from/to.

### DIFF
--- a/weepickle/test/src-jvm/com/rallyhealth/weepickle/v1/example/JvmPrimitiveTests.scala
+++ b/weepickle/test/src-jvm/com/rallyhealth/weepickle/v1/example/JvmPrimitiveTests.scala
@@ -16,7 +16,7 @@ object JvmPrimitiveTests extends TestSuite {
       test("LocalTime") { rw(LocalTime.parse("00:00"), "\"00:00\"") }
       test("LocalDateTime") { rw(LocalDateTime.parse("2000-01-01T00:00:00"), "\"2000-01-01T00:00:00\"") }
       test("OffsetDateTime") { rw(OffsetDateTime.parse("2007-12-03T10:15:30+01:00"), "\"2007-12-03T10:15:30+01:00\"") }
-      test("ZonedDateTime") {rw(ZonedDateTime.parse("2000-01-01T00:00:00Z"), "\"2000-01-01T00:00:00Z\"") }
+      test("ZonedDateTime") {rw(ZonedDateTime.parse("2000-01-01T00:00:00.000-06:00[America/Chicago]"), "\"2000-01-01T00:00:00.000-06:00[America/Chicago]\"") }
       test("Instant") {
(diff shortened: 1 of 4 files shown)


weepickle/test/src-jvm/com/rallyhealth/weepickle/v1/example/JvmPrimitiveTests.scala
```diff
@@ -16,6 +16,7 @@ object JvmPrimitiveTests extends TestSuite {
       test("LocalTime") { rw(LocalTime.parse("00:00"), "\"00:00\"") }
       test("LocalDateTime") { rw(LocalDateTime.parse("2000-01-01T00:00:00"), "\"2000-01-01T00:00:00\"") }
       test("OffsetDateTime") { rw(OffsetDateTime.parse("2007-12-03T10:15:30+01:00"), "\"2007-12-03T10:15:30+01:00\"") }
+      test("ZonedDateTime") {rw(ZonedDateTime.parse("2000-01-01T00:00:00Z"), "\"2000-01-01T00:00:00Z\"") }
```
b0bd: should we test an actual zone?

robert-j-miller: You mean one that isn't the same as UTC? We can I guess, but I don't see the difference.

dawna (approved):

weepickle/test/src-jvm/com/rallyhealth/weepickle/v1/example/JvmPrimitiveTests.scala
vito-c: pretty easy to add though? can  I suggest PST :D

jducoeur: Agreed -- UTC is kind of a special case, so I'd recommend adding one other zone as a belt-and-suspenders check.  Other than that, this looks good...

russellremple: But it wouldn't be PST, it would be something like "2000-01-01T00:00:00-08:00[America/Los_Angeles]"

robert-j-miller: Fair enough, swapped in Central time

vito-c (approved):

russellremple (approved):
lgtm

jducoeur (approved):

htmldoug (approved):
Thanks!

plokhotnyuk:
Beware, that ZonedDateTime.parse produces wrong ZoneOffset at DST fall with JDK 8: https://bugs.openjdk.java.net/browse/JDK-8066982

PR #18, opened 2020-04-09
Down the rabbit hole with auto-serialization for Enumerations

russellremple:
...if you go chasing enumerations, and you know you're going to fall...

(for posterity)

@htmldoug

### DIFF
--- a/implicits/src/com/rallyhealth/weepickle/v1/implicits/internal/Macros.scala
+++ b/implicits/src/com/rallyhealth/weepickle/v1/implicits/internal/Macros.scala
@@ -67,7 +67,9 @@ object Macros {
       assumeDefaultNone: Boolean,
       omitDefault: Boolean,
       default: c.Tree,
-      localTo: TermName,
+      localToName: TermName,
+      localToDef: Tree,
+      localFromDef: Tree,
       aggregate: TermName
     ) {
       def writingCheckDefault: Boolean = hasDefault && omitDefault
@@ -130,17 +132,36 @@ object Macros {
         // include .erasure to represent varargs as "Seq", not "Whatever*"
         val isOptionWithoutDefault = !isParamWithDefault &&
           argSym.typeSignature.erasure.typeConstructor.toString == "Option"
+        val argType = argTypeFromSignature(tpe, typeParams, argSym.typeSignature)
+
+        val isEnum = argType.baseClasses.exists(_.asClass.fullName == "scala.Enumeration.Value")
+
+        val localToDef =
+          if (isEnum) {
+            // brute force to get the outer class -- @todo maybe there's a better way...
+            val getOuter = c.parse(argType.toString.replace(".Value", ""))
+            q"implicitly[${c.prefix}.To[String]].map($getOuter.withName(_))"
+          } else
+            q"implicitly[${c.prefix}.To[$argType]]"
+
+        val localFromDef =
+          if (isEnum)
+            q"implicitly[${c.prefix}.From[String]].comap[$argType](_.toString)"
+          else
+            q"implicitly[${c.prefix}.From[$argType]]"
 
         new Argument(
           i = index,
           raw = argSym.name.toString,
           mapped = customKey(argSym).getOrElse(argSym.name.toString),
-          argType = argTypeFromSignature(tpe, typeParams, argSym.typeSignature),
+          argType,
           hasDefault = isParamWithDefault,
           assumeDefaultNone = isOptionWithoutDefault,
           omitDefault = shouldDropDefault(argSym),
           default = deriveDefault(companion, index, isParamWithDefault, isOptionWithoutDefault),
-          localTo = TermName("localTo" + index),
+          localToName = TermName("localTo" + index),
+          localToDef,
+          localFromDef,
           aggregate = TermName("aggregated" + index)
         )
       }
@@ -292,9 +313,9 @@ object Macros {
       if (args.size > 64) {
         c.abort(c.enclosingPosition, "weepickle does not support serializing case classes with >64 fields")
       }
-      q"""
+      val tree = q"""
         ..${for (arg <- args)
-        yield q"private[this] lazy val ${arg.localTo} = implicitly[${c.prefix}.To[${arg.argType}]]"}
+        yield q"private[this] lazy val ${arg.localToName} = ${arg.localToDef}"}
         new ${c.prefix}.CaseR[$targetType]{
           override def visitObject(length: Int) = new CaseObjectContext{
             ..${for (arg <- args)
@@ -342,11 +363,13 @@ object Macros {
             def subVisitor: com.rallyhealth.weepickle.v1.core.Visitor[_, _] = currentIndex match{
               case -1 => com.rallyhealth.weepickle.v1.core.NoOpVisitor
               case ..${for (arg <- args)
-        yield cq"${arg.i} => ${arg.localTo} "}
+        yield cq"${arg.i} => ${arg.localToName} "}
             }
           }
         }
       """
+      // println(s"Reading wrapCaseN tree = $tree")
+      tree
     }
     def mergeTrait(subtrees: Seq[Tree], subtypes: Seq[Type], targetType: c.Type): Tree = {
       q"${c.prefix}.To.merge[$targetType](..$subtrees)"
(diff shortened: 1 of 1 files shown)


vito-c:
don't go chasing enumerations stick to the case classes and objects that you're used to~

plokhotnyuk:
@russellremple `withName` is not the best option to resolve enum values by name. It uses iterator over an immutable bit set to scan through the set of enums and do a lot of boxing of `Int` values.

Bellow is a reverse flame graph that was built for results of the benchmark that parses an array of enums with weePickle using [the same method](https://github.com/plokhotnyuk/jsoniter-scala/blob/280cd170b23a1d0ccdac7fd481d16a2e21b8b3ef/jsoniter-scala-benchmark/src/main/scala/com/github/plokhotnyuk/jsoniter_scala/benchmark/WeePickleFromTos.scala#L45) to resolve names of the enum of 4 values.

image

Moreover, an overhead of boxing will increase if the size of the enum set will grow.

htmldoug:
Wow, I didn't realize `withName()` was so unoptimized. cc: @perryjp

implicits/src/com/rallyhealth/weepickle/v1/implicits/internal/Macros.scala
```diff
@@ -140,7 +140,9 @@ object Macros {
           if (isEnum) {
             // brute force to get the outer class -- @todo maybe there's a better way...
             val getOuter = c.parse(argType.toString.replace(".Value", ""))
-            q"implicitly[${c.prefix}.To[String]].map($getOuter.withName(_))"
+            q"""val nameToValue = scala.collection.mutable.Map.empty[String, $argType] ++ $getOuter.values.map(v => v.toString -> v)
+                implicitly[${c.prefix}.To[String]].map(nameToValue)
+             """
```
plokhotnyuk, line 156: Please consider to use `CHM` instead to lookup properly in case of dynamic extension of the enum set:
```
scala> object LocationType extends Enumeration { val GPS = Value(1, "GPS"); val IP = Value(2, "IP"); def more(name: String) = Value(nextId, name) }
defined object LocationType

scala> LocationType.values
res0: LocationType.ValueSet = LocationType.ValueSet(GPS, IP)

scala> LocationType.more("Postal Address")
res1: LocationType.Value = Postal Address

scala> LocationType.values
res2: LocationType.ValueSet = LocationType.ValueSet(GPS, IP, Postal Address)

scala> val enumCache = new java.util.concurrent.ConcurrentHashMap[String, LocationType.Value]()
enumCache: java.util.concurrent.ConcurrentHashMap[String,LocationType.Value] = {}

scala> def withName(name: String): LocationType.Value = enumCache.computeIfAbsent(name, LocationType.withName)
withName: (name: String)LocationType.Value

scala> withName("GPS")
res3: LocationType.Value = GPS

scala> LocationType.more("Address")
res4: LocationType.Value = Address

scala> withName("Address")
res5: LocationType.Value = Address

scala> withName("XXX")
java.util.NoSuchElementException: No value found for 'XXX'
  at scala.Enumeration.$anonfun$withName$2(Enumeration.scala:159)
  at scala.Enumeration.withName(Enumeration.scala:159)
  at .$anonfun$withName$1(<console>:1)
  at java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1660)
  at .withName(<console>:1)
  ... 28 elided
```

plokhotnyuk: Moreover, `java.util.concurrent.ConcurrentHashMap` stores hash codes separately so will return cached values faster than `scala.collection.mutable.HashMap`. Yet another `CHM`'s feature that the table for backets will be created on the first call of `computeIfAbsent`, not upfront.

russellremple: Oh, I like that, thanks!

russellremple: But defining an enumeration like that is a bit pathological... not really an enumeration anymore. If we assume a static set of values, it seems like pre-loading the values would be more performant (in the short run -- no diff amortized in the long run). I do like the idea that enumerations you never use don't take up extra memory.

htmldoug: If you never use them, I don't think the classloader would even load them.

plokhotnyuk:
Please see difference in results for [this](https://github.com/plokhotnyuk/jsoniter-scala/commit/c7b48d5ddab3c0973d24ea87806eb0901770c5c6) commit.

## Before
```
[info] Benchmark                             (size)   Mode  Cnt       Score      Error  Units
[info] ArrayOfEnumsReading.avSystemGenCodec     128  thrpt    5  125417.028 ± 7978.193  ops/s
[info] ArrayOfEnumsReading.borerJson            128  thrpt    5   92371.028 ± 3239.216  ops/s
[info] ArrayOfEnumsReading.jacksonScala         128  thrpt    5  137093.637 ± 4876.654  ops/s
[info] ArrayOfEnumsReading.sprayJson            128  thrpt    5   61673.755 ± 1210.588  ops/s
[info] ArrayOfEnumsReading.uPickle              128  thrpt    5  108992.872 ±  559.195  ops/s
[info] ArrayOfEnumsReading.weePickle            128  thrpt    5  119829.647 ± 1483.497  ops/s
```

## After
```
[info] Benchmark                             (size)   Mode  Cnt       Score       Error  Units
[info] ArrayOfEnumsReading.avSystemGenCodec     128  thrpt    5  147601.012 ± 11644.643  ops/s
[info] ArrayOfEnumsReading.borerJson            128  thrpt    5  155107.924 ±  8589.481  ops/s
[info] ArrayOfEnumsReading.jacksonScala         128  thrpt    5  164926.548 ±   928.293  ops/s
[info] ArrayOfEnumsReading.sprayJson            128  thrpt    5   76496.355 ±  1291.351  ops/s
[info] ArrayOfEnumsReading.uPickle              128  thrpt    5  144977.867 ±  1580.067  ops/s
[info] ArrayOfEnumsReading.weePickle            128  thrpt    5  143804.008 ±   917.731  ops/s
```

It happened that using `get` and `put` instead of `computeIfAbsent` shows better performance.

implicits/src/com/rallyhealth/weepickle/v1/implicits/internal/Macros.scala
russellremple: I meant "never use" in terms of the values encountered in the source data being deserialized (like Json), not code references.

htmldoug:
Questions:
1. Will this compose, e.g. with `Option`, or would we need a separate solution for that?
2. If I have a `FromTo[Suit.Value]` in scope, would that take precedence over the macro implementation?

implicits/src/com/rallyhealth/weepickle/v1/implicits/internal/Macros.scala
htmldoug:
```scala
import com.rallyhealth.weejson.v1.jackson.FromJson
import com.rallyhealth.weepickle.v1.WeePickle

object Suit extends Enumeration {

  val Spades = Value

  println("Suit initialized.")

  implicit val pickler = fromTo[String].bimap[Value](_.toString, withName)
}

case class Card(name: String, suit: Option[Suit.Value] = None)
object Card {
  implicit val pickler = WeePickle.macroFromTo[Card]
}

object Test extends App {

  val card = FromJson("""{"name": "Joker"}""").transform(ToScala[Card])

  println("Card with no Suit: " + card)
  println("-----------------------")
  println("Forcing classloading of Suit...")

  val forceClassloading = Suit.Spades
}
```

```
Card with no Suit: Card(Joker,None)
-----------------------
Forcing classloading of Suit...
Suit initialized.
```

russellremple:
As you expected, the answers are "no" and "no". That is the nice thing about implicits -- they compose. This is just for the common case where a data element is an enumeration (and not even an Enumeratum enumeration). So pretty narrow. And there is no precedence -- if it is an Enumeration, it gets this treatment, otherwise we look for an implicit. I wonder if the behavior could be defined _as_ an implicit rather than trying to go around implicit resolution, as long as it could be defined with a lower priority, then the answers would be "yes" and "yes". Thoughts?

htmldoug:
>  I wonder if the behavior could be defined as an implicit rather than trying to go around implicit resolution

That would fit in better with the library. The expectation is that when you define a custom type, you toss a `FromTo[T]` in the companion object. `Enumeration.Value` is a custom type and I'd expect to create a custom pickler. Haoyi also suggested something to that effect in https://github.com/lihaoyi/upickle/issues/147#issuecomment-213767487.

PR #42, opened 2020-08-22
Support case classes with > 64 fields.

htmldoug:
## Summary
The limitation of 64 fields came from a `var found: Long` used as a bitset to track which fields we've seen. I've changed the macro to add `var found1: Long`, `var found2: Long`, etc. as needed. Supports up to 2^31 fields now or whatever the jvm limit is.

## Motivation
@brianzhangrally has a response from an external API with 72 fields.

## Binary Compatibility
If I could break bin compat, I'd refactor `CaseR.found` and `CaseR.visitValue` into the macro and generalize, but it's a trait, so I can't touch it. Slightly smelly, but all the alternatives I can think of introduce code duplication and weaken cohesion.

@russellremple


weepickle/test/src/com/rallyhealth/weepickle/v1/FailureTests.scala
```diff
@@ -203,14 +203,14 @@ object FailureTests extends TestSuite {
...
       assert(FromJson(written63).transform(ToScala[Big63]) == b63)
       val written64 = FromScala(b64).transform(ToJson.string)
       assert(FromJson(written64).transform(ToScala[Big64]) == b64)
-      val err = compileError(
-        "{implicit val b64rw: com.rallyhealth.weepickle.v1.WeePickle.FromTo[Big65] = com.rallyhealth.weepickle.v1.WeePickle.macroFromTo}"
-      )
-      assert(err.msg.contains("weepickle does not support serializing case classes with >64 fields"))
+      val written65 = FromScala(b65).transform(ToJson.string)
+      assert(FromJson(written65).transform(ToScala[Big65]) == b65)
     }
```
danny-goldberg-rally: should this test case move from `FailureTests.scala` -> `MacroTests.scala`?

weepickle/test/src/com/rallyhealth/weepickle/v1/FailureTests.scala
```diff
@@ -203,14 +203,14 @@ object FailureTests extends TestSuite {
         com.rallyhealth.weepickle.v1.WeePickle.macroFromTo
       implicit val b64rw: com.rallyhealth.weepickle.v1.WeePickle.FromTo[Big64] =
         com.rallyhealth.weepickle.v1.WeePickle.macroFromTo
+      implicit val b65rw: com.rallyhealth.weepickle.v1.WeePickle.FromTo[Big65] =
+        com.rallyhealth.weepickle.v1.WeePickle.macroFromTo
       val written63 = FromScala(b63).transform(ToJson.string)
       assert(FromJson(written63).transform(ToScala[Big63]) == b63)
```
danny-goldberg-rally: I see where these are instantiated, but where are these `BigN` case classes defined?

implicits/src/com/rallyhealth/weepickle/v1/implicits/MacroImplicits.scala
```diff
@@ -329,15 +329,19 @@ object MacroImplicits {
...
-      if (args.size > 64) {
-        c.abort(c.enclosingPosition, "weepickle does not support serializing case classes with >64 fields")
+    def wrapCaseN(companion: c.Tree, args: Seq[Argument], targetType: c.Type, varargs: Boolean) = {
+
+      val addlBitSets = math.max(0, (args.size - 1) / 64) // number of additional bitsets, excluding `CaseR.found`
+
+      def foundBitSet(i: Int) = i match {
+        case 0 => TermName("found")
+        case i => TermName(s"found$i")
       }
```
russellremple, line 485: Is this fast? (specifically for the "normal" case with << 64 fields)

implicits/src/com/rallyhealth/weepickle/v1/implicits/MacroImplicits.scala
```diff
@@ -329,15 +329,19 @@ object MacroImplicits {
...
+        for (arg <- args if arg.hasDefault)
+          yield q"if (isMissing(${arg.i})) {setFound(${arg.i}); storeAggregatedValue(${arg.i}, ${arg.default})}"
       }"""
 
     override def wrapObject(t: Tree) = q"new ${c.prefix}.SingletonR($t)"
 
-    override def wrapCaseN(companion: Tree, args: Seq[Argument], targetType: Type, varargs: Boolean): Tree = {
-      if (args.size > 64) {
-        c.abort(c.enclosingPosition, "weepickle does not support serializing case classes with >64 fields")
+    def wrapCaseN(companion: c.Tree, args: Seq[Argument], targetType: c.Type, varargs: Boolean) = {
```
russellremple: return type?

russellremple (commented):
very cool

implicits/src/com/rallyhealth/weepickle/v1/implicits/MacroImplicits.scala
htmldoug: Code emitted for <= 64 fields is exactly the same as before. All the new stuff goes no-op, so perf should be identical.

I didn't even bother to bench > 64 since it's such a rare case.

implicits/src/com/rallyhealth/weepickle/v1/implicits/MacroImplicits.scala
htmldoug: no idea how that happened.

weepickle/test/src/com/rallyhealth/weepickle/v1/FailureTests.scala
htmldoug: yeah. done.

weepickle/test/src/com/rallyhealth/weepickle/v1/FailureTests.scala
htmldoug: At end of the file, I think.

jducoeur (approved):

PR #45, opened 2020-08-24
Fix Scaladoc errors/warnings.

htmldoug:


core/src/com/rallyhealth/weepickle/v1/core/Visitor.scala
```diff
@@ -21,7 +21,7 @@ import java.time.Instant
   *           a [[ArrVisitor]] and [[ObjVisitor]] via [[ObjArrVisitor.visitValue()]].
   *           For example, this might be a weejson.Str that gets passed into an
   *           [[ObjVisitor]] that's building up a weejson.Obj to be returned on [[ObjVisitor.visitEnd()]].
-  *           Often [[T]] will be the same type as [[J]] for visitors that return things,
+  *           Often `T` will be the same type as J` for visitors that return things,
```
danny-goldberg-rally, line 24:
```suggestion
  *           Often `T` will be the same type as `J` for visitors that return things,
```

htmldoug: Thanks. Fixed on v1.

PR #46, opened 2020-08-24
XML

russellremple:
Introduce limited XML support. Note that because XML parsing results in all strings, I changed the default Tos to allow visitString on non-string primitive types:boolean, int, short, and byte. Note that double and float were already doing that.
@htmldoug @jducoeur


weejson/xml/test/src/com/rallyhealth/weejson/v1/xml/TestApp.scala
```diff
@@ -0,0 +1,44 @@
...
+
+object TestApp extends App {
+  //
+  //  val xml = """---
+  //               |# A list of tasty fruits
+  //               |- Apple
+  //               |- Orange
+  //               |- Strawberry
+  //               |- Mango
+  //               |...""".stripMargin
```
htmldoug:
```suggestion
```

weepickle/test/src/com/rallyhealth/weepickle/v1/JsonPointerVisitorTests.scala
```diff
@@ -35,7 +35,7 @@ object JsonPointerVisitorTests extends TestSuite {
       test - assertPathFailure("""{"foo": ["", -666], "s": "", "i": 5, "b": true}""", "/foo/1")
       test - assertPathFailure("""{"foo": ["", -666, ""], "s": "", "i": 5, "b": true}""", "/foo/1")
       test - assertPathFailure("""{"foo": [], "s": -666, "i": 5, "b": true}""", "/s")
-      test - assertPathFailure("""{"foo": [], "s": "", "i": "-666", "b": true}""", "/i")
+      test - assertPathFailure("""{"foo": [], "s": "", "i": "-666NAN", "b": true}""", "/i")
```
htmldoug, line 38: What is this?!

htmldoug (approved):
`<!-- lgtm -->`

weepickle/test/src/com/rallyhealth/weepickle/v1/JsonPointerVisitorTests.scala
russellremple: because I changed DefaultTos for int to allow visitString, then this thing that used to fail now succeeds ("-666".toInt == -666). So "-666NAN" makes sure it actually fails ("-666NAN".toInt *boom*).

htmldoug: Got it, thanks!

jducoeur (approved):
Looks like a good first cut!  When this gets merged, I'll update the public discussion accordingly...

implicits/src/com/rallyhealth/weepickle/v1/implicits/DefaultTos.scala
```diff
@@ -48,6 +49,7 @@ trait DefaultTos extends com.rallyhealth.weepickle.v1.core.Types with Generated
 
   implicit val ToInt: To[Int] = new NumericTo[Int] {
     override def expectedMsg = "expected number"
+    override def visitString(cs: CharSequence) = cs.toString.toInt
```
htmldoug, line 52: What are the downsides of adding this type coercion for non-XML types, if any?

russellremple: It is additive, so it does not introduce any new cases where things that used to work now break. It certainly add cases where things that used to break now work (like `"-666"` as a string above). So I guess the downside is reduced static type safety relative to that, i.e., if you wanted some warning that you really ought to represent numbers as numbers instead of strings, you aren't going to get it anymore. And some strings will parse as ints and others won't, so all type conversion errors become runtime errors. HOWEVER, in nearly all cases (e.g., JSON parser) what `visit*` method to call is already inferred at runtime, so it isn't exactly "static" today, so there isn't much static type safety being giving up IMO. 

Also, as inferred in the PR description, adding this for boolean, int, short, and byte makes us _more_ consistent with what is already being done for double and float. If we want to fall back on maximum static safety, `ToDouble` and `ToFloat` should be modified to remove their implementations of `visitString`.

PR #50, opened 2020-10-07
Throw an exception when parsing zero tokens rather than returning null

AndyKirsch:
As discussed when writing code like 
`FromJson(path.toFile).transform(ToScala[Something])`
It would return null if the file was empty which was different from the behavior of `Json.parse(new FileInputStream(file))` which throws on an empty file.

I couldn't find a way to have this not change the behavior in the `FromJson(path.toFile).transform(ToScala[Option[Something]])` case and continue to return None though.

### DIFF
--- a/weejson/jackson/test/src/com/rallyhealth/weejson/v1/jackson/WeeJacksonSpec.scala
+++ b/weejson/jackson/test/src/com/rallyhealth/weejson/v1/jackson/WeeJacksonSpec.scala
@@ -1,7 +1,8 @@
 package com.rallyhealth.weejson.v1.jackson
 
-import java.io.StringWriter
+import java.io.{File, StringWriter}
 
+import com.fasterxml.jackson.core.io.JsonEOFException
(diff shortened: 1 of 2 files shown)


weejson/jackson/test/src/com/rallyhealth/weejson/v1/jackson/WeeJacksonSpec.scala
```diff
@@ -34,4 +35,13 @@ class WeeJacksonSpec
       }
     }
   }
+
+  "empty file input" - {
+    "to required" in {
+      val file = new File(getClass.getResource("/empty.json").getPath)
```
htmldoug, line 41: Where does `empty.json` reside?

htmldoug (approved):

htmldoug (approved):

jducoeur (approved):

arkban (approved):

PR #52, opened 2020-10-19
VisitorJsonGenerator.getOutputContent should not throw

htmldoug:
I found [something](https://github.com/elastic/elasticsearch/blob/3deebc280405ead358973f3e49e48395f39877ed/libs/x-content/src/main/java/org/elasticsearch/common/xcontent/json/JsonXContentGenerator.java#L454-L455) that actually uses this method. `null` is an acceptable value. Throwing is not.

### DIFF
--- a/weejson/jackson/src/com/rallyhealth/weejson/v1/jackson/VisitorJsonGenerator.scala
+++ b/weejson/jackson/src/com/rallyhealth/weejson/v1/jackson/VisitorJsonGenerator.scala
@@ -203,7 +203,7 @@ class VisitorJsonGenerator[J](
 
   override def writeTree(rootNode: TreeNode): Unit = objectCodec.writeValue(this, rootNode)
 
-  override def getOutputContext: JsonStreamContext = throw notSupported
+  override def getOutputContext: JsonStreamContext = null
 
   override def flush(): Unit = ()
 


weejson/jackson/src/com/rallyhealth/weejson/v1/jackson/VisitorJsonGenerator.scala
```diff
@@ -203,7 +203,7 @@ class VisitorJsonGenerator[J](
 
   override def writeTree(rootNode: TreeNode): Unit = objectCodec.writeValue(this, rootNode)
 
-  override def getOutputContext: JsonStreamContext = throw notSupported
+  override def getOutputContext: JsonStreamContext = null
```
crispywalrus, line 206: throwing is bad, yes. but `null` is as bad. Would `JsonStreamContext.empty` (so you get back the expected type but it doesn't really have a context data) be creatable?

htmldoug: I'd take a PR. `inRoot()`, `inObject()`, `inArray()` would all be possible to implement over `stack: mutable.ArrayStack[ObjArrVisitor[Any, _]]` as well.

PR #55, opened 2021-01-13
SimpleVisitor should not accept visitNull()

htmldoug:
This behavior seems undesirable:

```
import play.api.libs.json.Json

case class Foo(s: String)
object Foo {
  implicit val format = Json.format[Foo]
  implicit val rw = WeePickle.macroFromTo[Foo]
}
val json = "null"
Json.parse(json).validate[Foo] // JsError(List((/s,List(ValidationError(List(error.path.missing),WrappedArray())))))
FromJson(json).transform(ToScala[Foo]) // null ?!
```

### DIFF
--- a/core/src/com/rallyhealth/weepickle/v1/core/Types.scala
+++ b/core/src/com/rallyhealth/weepickle/v1/core/Types.scala
@@ -114,8 +114,8 @@ trait Types { types =>
   trait From[In] {
     def narrow[K] = this.asInstanceOf[From[K]]
     def transform[Out](in: In, out: Visitor[_, Out]): Out = {
-      if (in == null) out.visitNull()
-      else transform0(in, out)
+      if (in == null) throw new NullPointerException()
+      transform0(in, out)
     }
     def transform0[Out](in: In, out: Visitor[_, Out]): Out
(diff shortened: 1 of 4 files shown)


arkban (approved):
+1, make the easy things easy

core/src/com/rallyhealth/weepickle/v1/core/Types.scala
```diff
@@ -114,8 +114,8 @@ trait Types { types =>
   trait From[In] {
     def narrow[K] = this.asInstanceOf[From[K]]
     def transform[Out](in: In, out: Visitor[_, Out]): Out = {
-      if (in == null) out.visitNull()
-      else transform0(in, out)
+      if (in == null) throw new NullPointerException()
```
russellremple, line 117: I'm not sure I like this part. Why not just call `visitNull` and let the visitor decide if it can handle it? (Which, in the case of most visitors extending `SimpleVisitor`, would throw a more descriptive `Abort` indicating what was expected and not an NPE.) Wouldn't it be less surprising for `FromScala(null).transform(to[Option[String]])` to return `None` instead of NPE?

htmldoug:
Fixed by https://github.com/rallyhealth/weePickle/pull/77.

PR #56, opened 2021-01-22
Optimize PlayJson for mem usage.

htmldoug:
I'd like to be able to extend `PlayJson` to create a variant with a memoized `hashCode`. This will let me deduplicate nodes in the tree to save heap usage on duplication-heavy json texts.


russellremple (approved):
lgtm

jducoeur (approved):

Issue #68, opened 2021-04-20
Unable to import yaml in version 1.4.0


ashwinbhaskar:
From the examples, doing this import
```
import com.rallyhealth.weejson.v1.yaml.{FromYaml, ToYaml}
```
gives the error `object yaml is not a member of package com.rallyhealth.weejson.v1`

jducoeur:
Hmm.  Just guessing (I haven't tried using the YAML functionality myself), but it looks to me like that requires an additional dependency -- I would try:
```
libraryDependencies += "com.rallyhealth" %% "weeyaml-v1" % "version"
```

ashwinbhaskar:
@jducoeur yes, that did the work. Would you like me to update the readme with this?

jducoeur:
I'm not the primary author, but I suspect a PR would be welcomed -- thanks!

PR #69, opened 2021-04-20
Add weeyaml dependency and imports

ashwinbhaskar:


readme.md
```diff
@@ -29,7 +29,10 @@ weePickle combines some of the best parts of the serialization ecosystem.
 ## sbt
 ```scala
 resolvers += "Rally Health" at "https://dl.bintray.com/rallyhealth/maven"
-libraryDependencies += "com.rallyhealth" %% "weepickle-v1" % "version"
+libraryDependencies ++= Seq(
+  "com.rallyhealth" %% "weepickle-v1" % "version",
+  "com.rallyhealth" %% "weeyaml-v1" % "version"
+)
```
jducoeur: Could you move this new dependency info down to the section about YAML?  That's kind of the point of how this is all structured -- it's broken into lots of separate dependencies so that you only need to depend on what you actually need.  So `weeyaml` is only appropriate if you're doing yaml processing.

ashwinbhaskar: done

htmldoug:
Fixes #68.

Thanks!

PR #75, opened 2021-05-26
Slightly faster java enums

htmldoug:
@yuryrabiankou @russellremple @jducoeur

### DIFF
--- a/weepickle/src/main/scala/com/rallyhealth/weepickle/v1/LowPriorityImplicits.scala
+++ b/weepickle/src/main/scala/com/rallyhealth/weepickle/v1/LowPriorityImplicits.scala
@@ -22,7 +22,7 @@ abstract class LowPriorityImplicits
     var lastVset: E#ValueSet = null
 
     ToString.map { s =>
-      val vset = e.values // exploit that e.values returns a cached instance unless changed
+      val vset = e.values // exploit that e.values returns a cached instance unless new values are added
       if (vset ne lastVset) {
         // add any new values since last time
         vset.foreach(v => cache.put(v.toString, v))
--- a/weepickle/src/main/scala/com/rallyhealth/weepickle/v1/LowPriorityImplicits.scala
+++ b/weepickle/src/main/scala/com/rallyhealth/weepickle/v1/LowPriorityImplicits.scala
@@ -38,7 +38,7 @@ abstract class LowPriorityImplicits
   }
 
   private def fromEnumerationName[E <: scala.Enumeration](e: E): From[E#Value] = {
-    val cache = new ConcurrentHashMap[E#Value, String].asScala // mitigate withName() slowness.
+    val cache = new ConcurrentHashMap[E#Value, String].asScala // mitigate toString() slowness.
     FromString.comap(v => cache.getOrElseUpdate(v, v.toString))
   }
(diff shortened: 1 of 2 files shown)


weepickle/src/main/scala/com/rallyhealth/weepickle/v1/LowPriorityImplicits.scala
```diff
@@ -17,19 +19,27 @@ abstract class LowPriorityImplicits
 
   private def toEnumerationName[E <: scala.Enumeration](e: E): To[E#Value] = {
     val cache = new ConcurrentHashMap[String, E#Value] // mitigate withName() slowness.
+    var lastVset: E#ValueSet = null
 
     ToString.map { s =>
-      var enum = cache.get(s) // 68x faster than withName()
+      val vset = e.values // exploit that e.values returns a cached instance unless new values are added
+      if (vset ne lastVset) {
```
jducoeur, line 26: ... my initial reaction to this code was, "Wait -- Enumeration isn't immutable?!?"  Then I read the documentation.  Yet another reason for me to be unfond of `scala.Enumeration`.

jducoeur (approved):

weepickle/src/main/scala/com/rallyhealth/weepickle/v1/LowPriorityImplicits.scala
```diff
@@ -17,19 +19,27 @@ abstract class LowPriorityImplicits
 
   private def toEnumerationName[E <: scala.Enumeration](e: E): To[E#Value] = {
     val cache = new ConcurrentHashMap[String, E#Value] // mitigate withName() slowness.
+    var lastVset: E#ValueSet = null
```
arkban, line 22: Q: I thought I remember @jducoeur saying something about the default, is this form right or should it be this? 
```suggestion
    var lastVset: E#ValueSet = _
```

arkban (approved):

weepickle/src/main/scala/com/rallyhealth/weepickle/v1/LowPriorityImplicits.scala
htmldoug: Pretty sure it'd fail to compile. IIRC, only works for class fields. Also, it's getting phased out in scala 3 anyway.

arkban: > Also, it's getting phased out in scala 3 anyway.

That was the part I was thinking, I just couldn't remember if it was being enforced or dropped in Scala 3.

jducoeur: Yeah, the underscore initializer is sort of a horrible wart in the language.  In Scala 3 it becomes `uninitialized`.  In practice, it will wind up as `null` regardless, so calling a spade a spade isn't a terrible thing.

arkban: > In Scala 3 it becomes `uninitialized`

_That's_ what I couldn't remember! Thank you!

PR #77, opened 2021-06-02
Throw error on unexpected nulls for SimpleVisitor

marktoujiline:
Picking up where https://github.com/rallyhealth/weePickle/pull/55 left off.

This PR makes it so that any non-Option types that encounter `null` will throw. Reading `null`s will require setting the data type to `Option`.

The main point of contention I imagine is the asymmetric read/write this implementation proposes:
```scala
case class User(name: Option[String])
FromJson("""{"name": null}""").transform(ToScala[User]) ==> User(None)
FromScala(User(None)).transform(ToJson.string) ==> "{}"
```

I don't think it's a problem, but I'd be curious to know other opinions.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -262,7 +262,7 @@ Internally, jackson-core uses buffer pooling to achieve some of its performance.
 `FromJson` doesn't trust you and calls `close()` automatically after writing a single [json text](https://tools.ietf.org/html/rfc7159#section-2), which covers the vast majority of use cases. If you're working with multiple json texts separated by whitespace, jackson can handle it, but you have to drop down below the high level API and remember to close the parser/generator yourself.
 
 ## Value AST
-WeePickle includes its own AST named `Value`, largely unchanged from the upstream [uJson](http://www.lihaoyi.com/upickle/#uJson).
+WeePickle includes its own AST named `Value`, largely unchanged from the upstream [uJson](https://com-lihaoyi.github.io/upickle/#uJson).
 
 ```scala
 val obj = Obj(
@@ -289,11 +289,11 @@ FromJson("""{"foo":[42,"omg",true]}""").transform(Value) ==> obj
 ```
 
 See:
-- http://www.lihaoyi.com/upickle/#uJson
+- https://com-lihaoyi.github.io/upickle/#uJson
 - http://www.lihaoyi.com/post/uJsonfastflexibleandintuitiveJSONforScala.html
 
 ## MessagePack
-weePack is weePickle's [MessagePack](https://msgpack.org/index.html) implementation, largely unchanged from the upstream [uPack](http://www.lihaoyi.com/upickle/#uPack).
+weePack is weePickle's [MessagePack](https://msgpack.org/index.html) implementation, largely unchanged from the upstream [uPack](https://com-lihaoyi.github.io/upickle/#uPack).
 
 ### sbt
 ```scala
--- a/readme.md
+++ b/readme.md
@@ -293,19 +293,15 @@ See:
 - http://www.lihaoyi.com/post/uJsonfastflexibleandintuitiveJSONforScala.html
 
 ## Null Handling
+
 In JSON, `null` ["represents the intentional absence of any object value"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/null).
 This value is regularly used and must be supported.
 Scala also has a `null` value, but the usage is strongly discouraged, in part because it subverts the type system
 For example,
-```bash
-@ case class User(name: String)
-defined class User
-
-@ val user = User(null)
-user: User = User(null)
-
-@ user.name
-res6: String = null
+```scala
+case class User(name: String)
+val user = User(null)
+user.name // value is null
 ```
 The more equivalent value in Scala is `None`.
 Therefore, to support reading in JSON `nulls`, set the type to `Option`.
--- a/readme.md
+++ b/readme.md
@@ -292,6 +292,35 @@ See:
 - https://com-lihaoyi.github.io/upickle/#uJson
 - http://www.lihaoyi.com/post/uJsonfastflexibleandintuitiveJSONforScala.html
 
+## Null Handling
+In JSON, `null` ["represents the intentional absence of any object value"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/null).
+This value is regularly used and must be supported.
+Scala also has a `null` value, but the usage is strongly discouraged, in part because it subverts the type system
+For example,
+```bash
+@ case class User(name: String)
+defined class User
+
+@ val user = User(null)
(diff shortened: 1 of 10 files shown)


readme.md
```diff
@@ -289,11 +289,40 @@ FromJson("""{"foo":[42,"omg",true]}""").transform(Value) ==> obj
...
-- http://www.lihaoyi.com/upickle/#uJson
+- https://com-lihaoyi.github.io/upickle/#uJson
 - http://www.lihaoyi.com/post/uJsonfastflexibleandintuitiveJSONforScala.html
 
+## Null Handling
+In JSON, `null` ["represents the intentional absence of any object value"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/null).
+This value is regularly used and must be supported.
+Scala also has a `null` value, but the usage is strongly discouraged, in part because it subverts the type system
+For example,
+```bash
```
BurgersMcSlopshot: nit: language tag should be Scala, not BASH

weepickle-tests/src/test/scala/com/rallyhealth/weepickle/v1/TestUtil.scala
```diff
@@ -59,6 +59,13 @@ class TestUtil[Api <: com.rallyhealth.weepickle.v1.Api](val api: Api) {
     }
   }
 
+  def rwNull[T: api.To: api.From](t: T, s: String) = {
+    intercept[Exception] {
+      FromJson(s).transform(api.to[T])
+      api.from[T].transform(t, ToJson.string)
+    }
+  }
```
htmldoug: I see that this was relaxed from https://github.com/rallyhealth/weePickle/pull/55/files#diff-f94bae8c817f05636dca69fa1b3afe71fbd28f73176a2c529310917070306c3cR64-R71 so that only one of these has to throw. Did the stricter check fail?

weepickle-tests/src/test/scala/com/rallyhealth/weepickle/v1/DefaultsTests.scala
```diff
@@ -63,11 +63,9 @@ object StringTests extends TestSuite {
   implicit val pickler = WeePickle.macroFromTo[A]
 
   override val tests: Tests = Tests {
-    test("write default")(FromScala(A(null)).transform(ToJson.string) ==> """{"d":null}""")
```
htmldoug, line 66: What happened to these test cases? Would it make sense to update them to document and enforce the new behavior?

marktoujiline: I deleted these tests because they were testing scenarios where you were passing in a `null` to a case class, which doesn't seem like a use case worth supporting.

weepickle-tests/src/test/scala/com/rallyhealth/weepickle/v1/TestUtil.scala
htmldoug: Looks like the actual behavior is closer to:
```scala
  def rwNull[T: api.To: api.From](t: T, s: String) = {
    intercept[Exception] {
      FromJson(s).transform(api.to[T])
    }
    api.from[T].transform(t, ToJson.string)
  }
```

weepickle-tests/src/test/scala/com/rallyhealth/weepickle/v1/DefaultsTests.scala
htmldoug: Got it. I agree that using `null` isn't idiomatic/recommended usage.

However, there's history here (https://github.com/rallyhealth/weePickle/pull/43/files#r475258892). I'd prefer to keep these tests around to detect future behavior changes in this situation, particularly as we look toward scala 3 support and possibly borrowing macro code from upickle, which has different behavior here.

htmldoug:
> The main point of contention I imagine is the asymmetric read/write

We've already relaxed the asymmetry in other places for better pragmatism, so that part is fine with me. Symmetry is more important to uPickle, which has to transparently serialize task inputs/outputs from mill tasks. weePickle makes different tradeoffs based on our needs.

htmldoug (changes requested):
Let's tighten back down the test specifications, then LGTM. And thank you for picking up the torch on this!

weepickle-tests/src/test/scala/com/rallyhealth/weepickle/v1/TestUtil.scala
marktoujiline: Right, because right now JsonGeneratorVisitor implements `visitNull` to write null. We could have it throw like
```
override def visitNull(): JsonGenerator = throw new Abort("got null in Scala value")
```
which would fail Scala values that contain `null`. Otherwise, I think we have to leave the constraint the way you wrote and allow null values to be written, but not read.

weepickle-tests/src/test/scala/com/rallyhealth/weepickle/v1/DefaultsTests.scala
marktoujiline: I brought the tests back. This particular test is still being removed because it is identical to the "write null" test below.

htmldoug (approved):
thanks!

readme.md
```diff
@@ -289,11 +289,36 @@ FromJson("""{"foo":[42,"omg",true]}""").transform(Value) ==> obj
...
+1. The value was a JSON `null`
+1. The value was missing
+
+Writing a None to JSON will cause the field to be omitted, resulting in an asymmetric read/write.
+```scala
+case class User(name: Option[String])
+
+FromJson("""{"name": null}""").transform(ToScala[User]) ==> User(None)
+FromScala(User(None)).transform(ToJson.string) ==> "{}"
+```
```
russellremple, line 318: Slight aside, but this would be symmetric:
```scala
  FromJson("{}").transform(ToScala[User]) ==> User(None)
  FromScala(User(None)).transform(ToJson.string) ==> "{}"
```
So the issue (if you want to call it that) is that for each valid Scala encoding of data, there are potentially multiple ways that data could be encoded in JSON, but (thankfully) not the other way around (surjection). Perhaps a more trivial, but still relevant (if not even more nit-picky on my part) example would be JSON sources with different element spacing, like `{ }`, `{  }`, `{   }`, `{    }`, ... all mapping to `User(None)`. Though one might not think of these as "different" JSONs any more than we would think about `User( None )` as a "different" case class, when considered as unparsed string, `{}` does not equal `{    }`. But clearly `User(None) == User( None )` in a compiled program.

Anyway, your explanation is good and I have no issues with it, and certainly my comments here would add nothing to help the user understand null handling... but I just needed to get this whole symmetry vs. surjectiveness thing off my chest! :)

russellremple (approved):
I love this! lgtm

PR #79, opened 2021-06-17
Updated WeePickle readme.md around Defaults, Options, @dropDefault, and Null Handling

Outwise:
+ added IDE "bsp" dir to .gitignore; IDE specific files shouldn't be in repos
+ Overhauled Defaults, Options, @dropDefault, and Null Handling. Condensed sections, then expanded examples to show typical use cases.
+ some consistency tweaks, minor corrections

## TODO
- [ ] could use help confirming changes are comprehensive, 
- [ ] no missed copy/paste errors, 
- [ ] accurate assumptions on expected results


readme.md
```diff
@@ -115,81 +115,184 @@ Case classes are serialized using the `apply` and `unapply` methods on their com
...
-FromScala(Dflt2(42, 43, 0)).transform(ToJson.string)        ==> """{"k": 0}"""
+@dropDefault case class DropMultiDflts(i: Int = 42, j: Int = 43, k: Int = 44)
+object DropMultiDflts {
+  implicit val rw: FromTo[DropMultiDflts] = macroFromTo
+}
+
+FromScala(DropMultiDflts(i = 42, j = 43, k = 999)).transform(ToJson.string)   ==> """{"k": 999}"""
 ```
 
+If attempting to deserialize JSON to a field that has no defaults, and the JSON value is invalid (e.g. missing, or `null`), the operation will throw an Abort Exception.
```
htmldoug, line 179: Mostly correct. `null` handling is up to the `To`. Most will throw, but [Option](https://github.com/rallyhealth/weePickle/blob/v1/weepickle-implicits/src/main/scala/com/rallyhealth/weepickle/v1/implicits/DefaultTos.scala#L234), [Array](https://github.com/rallyhealth/weePickle/blob/v1/weepickle-implicits/src/main/scala/com/rallyhealth/weepickle/v1/implicits/DefaultTos.scala#L248), and others may override `visitNull`. The case class macro itself doesn't do any special null-handling.

readme.md
```diff
@@ -115,81 +115,184 @@ Case classes are serialized using the `apply` and `unapply` methods on their com
...
 ## Options
-`Option[T]` is unwrapped when the Option is `Some` ([rationale](differences.md#options)):
+
+If your JSON input field could be valid, `null`, or missing from the JSON, then use an `Option[T]` to catch these cases. An `Option` field will be unwrapped when serialized to JSON (unless excluded from the JSON via `@dropDefaults`), and wrapped when deserialized. ([rationale](differences.md#options)):
 
 ```scala
-case class Maybe1(i: Option[Int])
-object Maybe1 {
-  implicit val rw: FromTo[Maybe1] = macroFromTo
+case class DfltOpt(i: Option[Int]) // equivalent to 'i: Option[Int] = None` for weepickling
```
htmldoug, line 190: nice clarification.

readme.md
```diff
@@ -115,81 +115,184 @@ Case classes are serialized using the `apply` and `unapply` methods on their com
...
+FromScala(DropMultiDfltOpt(i = Some(42), j = None, k = Some(999), l = None)).transform(ToJson.string)   ==> """{"j": null, "k": 999}"""
+// matching defaults and values are dropped: 
+// i(42 == 42), j(42 != None), k(999 != None), l(None == None) 
 ```
 
-But `Option` types are a special case where `None` is an assumed default if a default is not provided explicitly.
-So putting `@dropDefault` at the class level will apply to all `Option` types in the class, whether a default is provided explicitly or not.
+### Explicit Defaults
+
+An `Option` will have an [implicit default of `None`](#Nones-&-Nulls), unless you provide an explicit default ([Defaults](#Defaults)):
```
htmldoug, line 262: Broken links here.

readme.md
```diff
@@ -115,81 +115,184 @@ Case classes are serialized using the `apply` and `unapply` methods on their com
...
+}
+
+FromScala(DropMultiDflts(i = 42, j = 43, k = 999)).transform(ToJson.string)   ==> """{"k": 999}"""
 ```
 
+If attempting to deserialize JSON to a field that has no defaults, and the JSON value is invalid (e.g. missing, or `null`), the operation will throw an Abort Exception.
+
+### Option Defaults
+
+`Option` works the same way, except there is always an assumed implicit `None` default, if an explicit default is not provided. See [Option Defaults & Nulls](#Option-Defaults-&-Nulls).
```
htmldoug, line 183: Broken link here.

readme.md
```diff
@@ -115,81 +115,184 @@ Case classes are serialized using the `apply` and `unapply` methods on their com
 Anything else is not supported by default, but you can add support using Custom Picklers.
 
 ## Defaults
-If a field is missing upon deserialization, weePickle uses the default value if one exists.
+If a field is missing upon **deserialization**, weePickle uses the default value if one exists:
```
htmldoug, line 134: I know `deserialization` and `serialization` were already here, but they're a bit misleading. All of these behaviors have to do with case classes rather than serialization formats. For example, you can do `FromJson("{}").transform(ToMsgPack.string)`, which is both deserialization and serialization, but none of this will apply.

More README surgery would be welcome if interested. Also, totally fine to leave for a separate PR.

.gitignore
```diff
@@ -3,3 +3,5 @@ out/
 *.iml
 .idea
 .DS_Store
+# cloning the repo generates an IDE specific file, `.bsp/.sbt.json`. (https://users.scala-lang.org/t/solved-should-i-add-bsp-directory-in-gitignore/6921)
+.bsp%
```
htmldoug:
```suggestion
.bsp
```

htmldoug (approved):
Thanks so much for sprucing up the docs!

jducoeur (approved):

PR #81, opened 2021-07-12
Getting started with Scala 3

russellremple:
Only have common code between Scala 2 and 3 when Scala 2 libs can maintain binary compatibility. Any non-passing tests are commented out in the scala-3 source paths (same as in uPickle, e.g., AdvancedTests). Macros are based on uPickle macros with WeePickle-specific features like custom discriminator names, default-dropping behavior, special handling for Option defaults, and "Nullable" To's. 

There are still many TODOs, no benchmarks, etc. So this PR is really just the beginning.


weepickle-core/src/main/scala/com/rallyhealth/weepickle/v1/core/JsonPointerVisitor.scala
```diff
@@ -37,7 +37,7 @@ object JsonPointerVisitor {
     * Compared to something like a List[String] or List[Object], this does not require
     * extra String allocation or boxing unless we actually ask for the path.
     */
-  private trait HasPath {
+  trait HasPath { // Scala 3 needs this to be public -- doesn't break bin compat
```
AndyKirsch: Why does scala 3 need this?

russellremple: I'm actually surprised that Scala 2 let the private class leak out in a public constructor, but it did. And this is one place where the weird rules of bin compat work in our favor: technically changing something from private to public should affect bin compat, but Scala generates these as public interfaces in the bytecode, so from Mima perspective, it was already public. (Which is why, I think, it is hard to lock Scala bin compat down even by making everything private, because the compiler often winds up exposing private things as public.)
```
...JsonPointerVisitor.scala:97:32 
[error] 97 |private class JsonPointerVisitor[T, J](
[error]    |                                ^
[error]    |non-private constructor JsonPointerVisitor in class JsonPointerVisitor refers to private trait HasPath
[error]    |in its type signature [T, J]
[error]    |  (delegate: com.rallyhealth.weepickle.v1.core.Visitor[T, J], parentPath: 
[error]    |    com.rallyhealth.weepickle.v1.core.JsonPointerVisitor.HasPath
[error]    |  ): com.rallyhealth.weepickle.v1.core.JsonPointerVisitor[T, J]
```

russellremple: On second thought... TIL that you can mark the default constructor as private, which is apparently not the default for a private class (weird that it isn't -- seems like a bug). Anyway, this allows `HasPath` to remain private and makes both Scala 2 and 3 happy. See latest commit.

weejson/src/main/scala/com/rallyhealth/weejson/v1/Value.scala
```diff
@@ -215,7 +215,7 @@ object Obj {
   implicit def from(items: TraversableOnce[(String, Value)]): Obj = {
     val initialCapacity = items match {
       case is: mutable.IndexedSeq[_] => is.size
-      case _ => 2
+      case _                         => 2
```
BurgersMcSlopshot: not that it's going to affect anything, but this whitespace change is a bit nutty.

weepickle-implicits/src/main/scala/com/rallyhealth/weepickle/v1/implicits/DefaultFroms.scala
```diff
@@ -158,12 +153,12 @@ trait DefaultFroms
...
   * This needs to be split into a separate trait due to https://github.com/scala/bug/issues/11768
   */
-trait LowPriFroms extends com.rallyhealth.weepickle.v1.core.Types {
+trait LowPriFroms { this: Types =>
   implicit def SeqLikeFrom[C[_] <: Iterable[_], T](implicit r: From[T]): From[C[T]] =
     new From[C[T]] {
       def transform0[R](v: C[T], out: Visitor[_, R]): R = {
         val ctx = out.visitArray(v.size).narrow
-        v.asInstanceOf[Iterable[T]].foreach { t =>
+        v.iterator.foreach { t =>
```
htmldoug:
```suggestion
        v..asInstanceOf[Iterable[T]].foreach { t =>
```
seems to work as is. `Iterable.foreach` might be optimized.

russellremple: Done. I just did a "clean up" commit to remove unnecessary changes like these. Also unnecessary reformatting. Also brought back`def`s instead of `val`s in `RW` definitions, since recursion issues are addressed.

weejson/src/main/scala/com/rallyhealth/weejson/v1/Value.scala
russellremple: removed

weejson-play/src/main/scala/com/rallyhealth/weejson/v1/play/PlayJson.scala
```diff
@@ -41,9 +41,9 @@ class PlayJson extends com.rallyhealth.weejson.v1.AstTransformer[JsValue] {
 
   def visitNull(): JsValue = JsNull
 
-  val visitFalse: JsValue = JsBoolean(false)
+  def visitFalse(): JsValue = JsBoolean(false)
```
htmldoug:
```suggestion
  private val jsFalse = JsBoolean(false)
  def visitFalse(): JsValue = jsFalse
```

There are still some play 2.5 consumers.

https://github.com/playframework/playframework/blob/df6d51b1fa0ecd95a4a5d3e45fd83aec5eea6ab4/framework/src/play-json/src/main/scala/play/api/libs/json/JsValue.scala#L36

weejson-play/src/main/scala/com/rallyhealth/weejson/v1/play/PlayJson.scala
```diff
@@ -41,9 +41,9 @@ class PlayJson extends com.rallyhealth.weejson.v1.AstTransformer[JsValue] {
...
 
-  val visitTrue: JsValue = JsBoolean(true)
+  def visitTrue(): JsValue = JsBoolean(true)
```
htmldoug:
```suggestion
  private val jsTrue = JsBoolean(true)
  def visitTrue(): JsValue = jsTrue
```

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassFrom.scala
```diff
@@ -0,0 +1,118 @@
...
+
+import compiletime.{summonInline}
+import deriving.Mirror
+import scala.reflect.ClassTag
+import com.rallyhealth.weepickle.v1.core.{Annotator, ObjVisitor, Types, Visitor}
+
+trait CaseClassFromPiece extends MacrosCommon:
+  this: Types with Annotator =>
+
+  class CaseClassFrom[V](
```
htmldoug:
```suggestion
  private class CaseClassFrom[V](
```

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassTo.scala
```diff
@@ -0,0 +1,197 @@
...
+trait CaseClassToPiece extends MacrosCommon :
+  this: Types with Annotator =>
+
+  // must duplicate ToString definition here to avoid circular reference with DefaultTos
+  implicit val stringVisitor: To[String] = new SimpleTo[String] {
+    override def expectedMsg = "expected string"
+    override def visitString(cs: CharSequence): String = cs.toString
+  }
+
+  class CaseClassTo[T](
```
htmldoug:
```suggestion
  private class CaseClassTo[T](
```

weepickle-core/src/main/scala/com/rallyhealth/weepickle/v1/core/Types.scala
```diff
@@ -217,6 +213,40 @@ trait Types { types =>
...
+          "missing keys in dictionary: " + keys.mkString(", ")
+        )
+      }
+      protected def checkErrorMissingKeys(rawArgsLength: Int) = {
+        var bits = 0
+        for (v <- found) bits += java.lang.Long.bitCount(v)
+        bits != rawArgsLength
+      }
+    }
+    //+end-2
```
htmldoug: Deletable?
```suggestion
```

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassTo.scala
```diff
@@ -0,0 +1,197 @@
...
+    fromProduct: Product => T,
+    // parallel arrays in field definition order
+    fieldNames: Array[String],
+    defaultValues: Array[Option[() => AnyRef]],
+    createVisitors: => Array[Visitor[_, _]]
+  ) extends CaseR[T] :
+
+    private lazy val visitors = createVisitors
+    private val outOfBounds = -1
```
htmldoug:
```suggestion
    private inline val outOfBounds = -1
```

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassTo.scala
```diff
@@ -0,0 +1,197 @@
...
+      makeWithoutDefaults(elements)
+    end makeWithDefaults
+
+    private def makeWithoutDefaults(elements: Array[AnyRef]): T =
+      fromProduct(new Product {
+        def canEqual(that: Any): Boolean = true
+        def productArity: Int = arity
+        def productElement(i: Int): Any = elements(i)
+      })
+    end makeWithoutDefaults
```
htmldoug, line 68: nice optimization

weejson-play/src/main/scala/com/rallyhealth/weejson/v1/play/PlayJson.scala
russellremple: Yeah, I was going to do that and then thought "isn't that what mjson did that caused the memory leak?" and got spooked. Happy to make the change if you think it is safe. Let me know...

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassTo.scala
```diff
@@ -0,0 +1,197 @@
...
+
+      val reader = new CaseClassTo[T](
+        m.fromProduct,
+        fieldNames,
+        defaultValues,
+        createVisitors
+      )
+
+      val (isSealed, discriminator) = macros.isMemberOfSealedHierarchy[T]
+      if isSealed then annotate(reader, discriminator.getOrElse("$type"), macros.fullClassName[T]._1)
```
htmldoug: Do we need to use `AttributeTagged.tagName` here instead?

weejson-play/src/main/scala/com/rallyhealth/weejson/v1/play/PlayJson.scala
htmldoug: Yes, please. `JsBoolean` is immutable, so nothing spooky here.

For reference: mjson objects can't be shared because of https://github.com/bolerio/mjson/blob/e7a4da2daa6e17a63ec057948bc30818e8f44686/src/java/mjson/Json.java#L1928-L1941

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassTo.scala
russellremple: I can make the class private, but the constructor needs to remain public (which is the default), otherwise
```
[error] 94 |      val fromCaseClass = CaseClassFrom[T](
[error]    |                          ^^^^^^^^^^^^^
[error]    |Implementation restriction: cannot use private constructors in inlineinline methods
```

weepickle-core/src/main/scala/com/rallyhealth/weepickle/v1/core/Types.scala
russellremple: Yeah, I didn't wind up using it. It came from latest upickle. I'll delete it and we can add it in later if we need it.

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassTo.scala
russellremple: oooo... why?

htmldoug: It's used as a constant. As written, it creates a 4-byte field on every instance. [Depending on alignment](https://shipilev.net/jvm/anatomy-quarks/24-object-alignment/), JVM may use more heap. Admittedly, there won't be nearly as many of these instances as something like `ObjVisitor`, but free savings are free savings.

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassTo.scala
htmldoug: Will we be able to change the class later without breaking bin compat?

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassTo.scala
russellremple: Ouch! Yes, but it is in a very different place in the inheritance hierarchy. Playing around if it can be abstract in `MacrosCommon`.

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassTo.scala
russellremple: idk, and update: even making the class private causes issues (not sure why I didn't see it earlier -- maybe I forgot to `+clean`):
```
[error] -- Error: /Users/russell-remple/IdeaProjects/weePickle/bench/src/main/scala/bench/ScalaVersionBench.scala:70:36 
[error] 70 |    implicit val rw: FromTo[ADT0] = macroFromTo
[error]    |                                    ^^^^^^^^^^^
[error]    |undefined: new CaseClassToPiece_this.CaseClassTo[bench.ScalaVersionBench.ADT0] # -1: TermRef(AppliedType(TypeRef(TermRef(NoPrefix,val CaseClassToPiece_this),CaseClassTo),List(TypeRef(ThisType(TypeRef(ThisType(TypeRef(NoPrefix,module class bench)),module class ScalaVersionBench$)),class ADT0))),<init>) at inlining
[error]    | This location contains code that was inlined from CaseClassTo.scala:110
```

htmldoug: Would it be possible to make it an anonymous class?

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassTo.scala
russellremple: Making it abstract in `MacrosCommon` worked, but actually it makes a lot more sense to put it in the new `Annotator` trait, which both`CaseClassFromPiece` and `CaseClassToPiece` reference in their self types and `AttributeTagged` (now) implements.

weejson-play/src/main/scala/com/rallyhealth/weejson/v1/play/PlayJson.scala
russellremple: done

weejson-play/src/main/scala/com/rallyhealth/weejson/v1/play/PlayJson.scala
russellremple: done

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassFrom.scala
russellremple: done (made anonymous)

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassTo.scala
russellremple: done

weepickle-core/src/main/scala/com/rallyhealth/weepickle/v1/core/Types.scala
russellremple: done

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassTo.scala
russellremple: done

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassTo.scala
russellremple: thanks!

htmldoug (approved):

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/macros.scala
```diff
@@ -0,0 +1,181 @@
+package com.rallyhealth.weepickle.v1.implicits.macros
+
+import scala.quoted.{ given, _ }
+import deriving._, compiletime._
+
+inline def getDefaultParams[T]: String => Option[() => AnyRef] = ${ getDefaultParmasImpl[T] }
+def getDefaultParmasImpl[T](using Quotes, Type[T]): Expr[String => Option[() => AnyRef]] =
```
jducoeur, line 7: Minor detail (currently catching up), but while I notice it: `getDefaultParmasImpl` -> `getDefaultParamsImpl `.

PR #82, opened 2021-07-13
Fix: "match may not be exhaustive" scala 2.13 -Xlint warnings

htmldoug:
Test and workaround for https://github.com/com-lihaoyi/upickle/issues/345.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -205,3 +205,9 @@ lazy val weexml = project
       "com.fasterxml.jackson.dataformat" % "jackson-dataformat-xml" % "2.12.3",
     )
   )
+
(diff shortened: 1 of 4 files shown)


russellremple (approved):
lgtm

jducoeur (approved):
Yeah, that makes sense -- I can see where the error was coming out of this.

(Does lead me to wonder if `currentIndex` should theoretically be a refined type, but it's almost certainly not worth the effort and overhead.)

htmldoug:
I thought about  eliminating the second one by having the macro generate code for each field check rather than looping which may perform slightly better too, but that branch only executes over the exceptional case when fields are missing, so I decided that approach wasn't worth it either.

PR #83, opened 2021-07-13
Add `@implicitNotFound` for From/To/FromTo

### LINKED ISSUE #62: Improve compiler errors by annotating To/From with @implicitNotFound
It's becoming evident that new users don't understand implicit resolution msgs:

Screenshot 2021-01-30 at 2

We can do a bit better by annotating To/From with something like
```
@implicitNotFound("Could not find an implicit To[${T}]. Consider adding one with `object ${T} { implicit val to${T}: WeePickle.To[${T}] = macroTo }"
```

htmldoug:
Fixes #62.

cc: @sshankhdhar @b0bd


russellremple (approved):
lgtm

jducoeur (approved):

PR #85, opened 2021-07-15
Update uPickle benchmarks

htmldoug:
Quick and dirty evaluation of the perf improvements in https://github.com/com-lihaoyi/upickle/pull/338. It's a big improvement!

Jackson-core provides error paths out of the box. They're off by default in uPickle, but available now as opt-in. I'm measuring both ways so we can compare peak perf and apples-to-apples.

### DIFF
--- a/bench/src/main/scala/bench/JsonBench.scala
+++ b/bench/src/main/scala/bench/JsonBench.scala
@@ -1,33 +1,28 @@
 package bench
 
-import java.util.concurrent.TimeUnit
-
 import com.rallyhealth.weejson.v1.jackson.{FromJson, ToJson}
 import com.rallyhealth.weepickle.v1.Common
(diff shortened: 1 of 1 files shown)


plokhotnyuk:
[Here](https://plokhotnyuk.github.io/jsoniter-scala/) are results of more realistic benchmarks (uPickle and weePickle included!) that use real-world message samples and different data-types for parsing immediatelly from buffered bytes to data structures and serializing back.

htmldoug:
Thanks @plokhotnyuk! I wasn't sure if you'd pulled in the upickle update and rebenchmarked. I should have known you'd be on top of it!

Looks like jackson-core is still a bit faster than the refreshed uPickle JSON parser over most of your real world tests, even without accounting for error path tracking. Great to know that it's still keeping up.

PR #91, opened 2021-08-13
1st class BufferedValue

russellremple:
Generally speaking, overusing `Value` is an anti-pattern... but it is often the most practical option in client code. But `Value` suffers from issues that `BufferedValue` does not: `BufferedValue` is a more faithful representation of the types that can be visited (not just JSON-specific types/encoding), the underlying structures are immutable, and it is generally faster (especially for numeric types). But `BufferedValue` is mostly an "internal" structure, used primarily for buffering the serialization of sealed traits and such. Here I'm attempting to add to `BufferedValue` the same features that we have in Value, so it can be used more readily by client code. (This more or less completes the work started in #86.) The goal is something as close to a drop-in replacement for `Value` as possible, within the constraints of binary compatibility.


weejson/src/main/scala/com/rallyhealth/weejson/v1/BufferedValue.scala
```diff
@@ -287,5 +327,31 @@ object BufferedValue extends Transformer[BufferedValue] {
...
+      }
+    )
+
+    override def visitUInt64(ul: Long): BufferedValue = visitFloat64StringParts(
+      cs = java.lang.Long.toUnsignedString(ul),
+      decIndex = -1,
+      expIndex = -1
+    )
+
+    override def visitInt32(i: Int): BufferedValue = visitInt64(i)
```
htmldoug:
```suggestion
    override def visitInt32(i: Int): BufferedValue = NumLong(i)
```

weejson/src/main/scala/com/rallyhealth/weejson/v1/BufferedValue.scala
```diff
@@ -287,5 +327,31 @@ object BufferedValue extends Transformer[BufferedValue] {
...
+
+    override def visitFloat32(f: Float): BufferedValue = visitFloat64(f)
```
htmldoug:
```suggestion
    override def visitFloat32(f: Float): BufferedValue = NumDouble(f)
```

weepickle/src/main/scala/com/rallyhealth/weepickle/v1/JsFromTos.scala
```diff
@@ -35,4 +35,58 @@ trait FromToValue extends MacroImplicits { this: Types with Annotator =>
...
+    ToBufferedValue.narrow[com.rallyhealth.weejson.v1.BufferedValue.Ext]
+  implicit def ToBufferedValueTimestamp: To[com.rallyhealth.weejson.v1.BufferedValue.Timestamp] =
+    ToBufferedValue.narrow[com.rallyhealth.weejson.v1.BufferedValue.Timestamp]
+  implicit def ToBufferedValueBool: To[com.rallyhealth.weejson.v1.BufferedValue.Bool] = ToBufferedValue.narrow[com.rallyhealth.weejson.v1.BufferedValue.Bool]
+  implicit def ToBufferedValueTrue: To[com.rallyhealth.weejson.v1.BufferedValue.True.type] =
+    ToBufferedValue.narrow[com.rallyhealth.weejson.v1.BufferedValue.True.type]
+  implicit def ToBufferedValueFalse: To[com.rallyhealth.weejson.v1.BufferedValue.False.type] =
+    ToBufferedValue.narrow[com.rallyhealth.weejson.v1.BufferedValue.False.type]
+  implicit def ToBufferedValueNull: To[com.rallyhealth.weejson.v1.BufferedValue.Null.type] =
+    ToBufferedValue.narrow[com.rallyhealth.weejson.v1.BufferedValue.Null.type]
```
htmldoug, line 64: Haven't tested, but would this work?

```suggestion
  val ToValue: To[com.rallyhealth.weejson.v1.Value] =
    new To.Delegate(com.rallyhealth.weejson.v1.Value)

  implicit def ToValueObj[T <: com.rallyhealth.weejson.v1.Value]: To[T] = ToValue.narrow[T]
```

htmldoug: Oh, it trips up `compileError("implicitly[com.rallyhealth.weepickle.v1.WeePickle.To[Nothing]]")`. Nevermind!

htmldoug (approved):

PR #92, opened 2021-09-03
Change parseLong(): len => end

htmldoug:
https://github.com/com-lihaoyi/upickle/pull/361 made me realize that the third param of `Util.parseLong` is intended to be `end`, not `len`. I didn't realize this assumption within weepickle because we don't use `Util.parseLong` for JSON parsing like upickle does.

I'm updating weepickle to be consistent to avoid any latent bugs and keep it simple to take future patches from upickle.

### DIFF
--- a/weepickle-core/src/main/scala/com/rallyhealth/weepickle/v1/core/Util.scala
+++ b/weepickle-core/src/main/scala/com/rallyhealth/weepickle/v1/core/Util.scala
@@ -43,16 +43,15 @@ object Util {
 
     intPortion + decPortion
   }
-  def parseLong(cs: CharSequence, start: Int, len: Int): Long = {
+  def parseLong(cs: CharSequence, start: Int, end: Int): Long = {
     // we store the inverse of the positive sum, to ensure we don't
     // incorrectly overflow on Long.MinValue. for positive numbers
     // this inverse sum will be inverted before being returned.
     var inverseSum: Long = 0L
     var inverseSign: Long = -1L
     var i: Int = start
-    val end = start + len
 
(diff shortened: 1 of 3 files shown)


weepickle-core/src/main/scala/com/rallyhealth/weepickle/v1/core/Util.scala
```diff
@@ -43,16 +43,15 @@ object Util {
...
     // we store the inverse of the positive sum, to ensure we don't
     // incorrectly overflow on Long.MinValue. for positive numbers
     // this inverse sum will be inverted before being returned.
     var inverseSum: Long = 0L
     var inverseSign: Long = -1L
     var i: Int = start
-    val end = start + len
 
-    if ((start | len | (cs.length - end) | end) < 0) throw new IndexOutOfBoundsException
+    if ((start | end | end - start | cs.length - end) < 0) throw new IndexOutOfBoundsException
```
jducoeur, line 54: Wow, hadn't noticed this bit before.  I can see the optimization, but boy I'm unused to thinking in terms of this sort of bit-twiddling these days.

jducoeur (approved):

weepickle-core/src/main/scala/com/rallyhealth/weepickle/v1/core/Util.scala
htmldoug: cats is pretty far removed from it. :)

I stole the approach from the java standard library where it's used heavily in `java.io.OutputStream` classes, etc.

russellremple (approved):
("lgtm", 0, 4)

PR #94, opened 2021-10-01
Scala 3 derive enum support

russellremple:
Derive special `From`, `To`, and `FromTo` type classes for string encoding of enums.


weepickle/src/main/scala-3/com/rallyhealth/weepickle/v1/LowPriorityImplicits.scala
```diff
@@ -40,16 +40,31 @@ abstract class LowPriorityImplicits extends AttributeTagged {
...
 
-  def fromToEnumerationId[E <: scala.Enumeration, V <: scala.Enumeration#Value](e: E)(implicit ev: V =:= e.Value): FromTo[V] = {
+  def fromToEnumerationId[E <: scala.Enumeration, V <: scala.Enumeration#Value](e: E)(
+    implicit ev: V =:= e.Value): FromTo[V] = {
     FromTo.join(ToInt.map(e.apply(_).asInstanceOf[V]), FromInt.comap(_.id))
   }
+
+  //  Native Scala 3 enums extend scala.reflect.Enum transparently.
+  //  By name encoding provided in default derives, so only id encoding provided here.
+  // TODO: do we even need this? If we do, should it be here (it isn't implicit)?
```
russellremple: What do you think @htmldoug ?

htmldoug: I only see one place that `fromToEnumerationId` is used in our current application code and I'm not sure it's even intentional. I'd be fine with leaving this out. Users can always handle with custom picklers.

weepickle-tests/src/test/scala-3/com/rallyhealth/weepickle/v1/EnumV3Tests.scala
```diff
@@ -0,0 +1,114 @@
...
+    case Horseshoe extends Charm("purple", 1983)
+    case Balloon   extends Charm("red", 1989)
+    case Tree      extends Charm("green", 1991)
+  }
+
+  def tests = Tests {
+    test("name") {
+      test("happy path - Fruits (raw enums)") - repeatedly {
+        FromJson(""""Peach"""").transform(ToScala[Fruit]) ==> Fruit.Peach
+        FromScala(Fruit.Peach).transform(ToJson.string) ==> """"Peach""""
```
htmldoug, line 43: @vito-c @mdotson Peach problem fixed!

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassFrom.scala
```diff
@@ -108,11 +109,28 @@ trait CaseClassFromPiece extends MacrosCommon:
       From.merge[T](writers:_*)
   }
 
+  /*
+   * Default string encoding for Scala 3 enums. Mirror should always be for the sealed trait generated by the
+   * enum, which is a Sum type. So any calls here with Product type mirrors are rejected as compilation errors.
+   */
+  inline def macroEnumFrom[T: ClassTag](using m: Mirror.Of[T]): From[T] = inline m match {
```
htmldoug, line 116: If this is only used internally, can it be made private?

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassTo.scala
```diff
@@ -124,6 +125,22 @@ trait CaseClassToPiece extends MacrosCommon :
...
+  /*
+   * Default string decoding for Scala 3 enums. Mirror should always be for the sealed trait generated by the
+   * enum, which is a Sum type. So any calls here with Product type mirrors are rejected as compilation errors.
+   */
+  inline def macroEnumTo[T](using m: Mirror.Of[T]): To[T] = inline m match {
+    case m: Mirror.ProductOf[T] =>
+      error("Enumeration macro used for non-sealed trait")
+
+    case m: Mirror.SumOf[T] =>
+      val valueOf: String => T = macros.enumValueOf[T]
```
htmldoug, line 137: Is there runtime cost for creating the `String => T` function?

htmldoug (approved):
The macro stuff is black magic to me, but the end result is fantastic.

weepickle/src/main/scala-3/com/rallyhealth/weepickle/v1/LowPriorityImplicits.scala
russellremple: That's kind of where my head was at too. I'll remove it and change the unit test to construct it locally.

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassTo.scala
russellremple: compile time :)

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/CaseClassFrom.scala
russellremple: I thought that, because it was called from `MacroImplicits` in the same package, I could make it package private. (Same with `macroEnumTo` in `CaseClassTo`.) But I was wrong. I guess there are different visibility rules for inline because each inline is resolved in local scope. So it really can't be private to some other scope per se.

htmldoug: Sounds good. Wiggle room for the future.

PR #95, opened 2021-10-13
Add Ops

htmldoug:
This is a first step toward open sourcing some of our higher-level weePickle code, starting with `ConcatFromInputs`. The big decision is where it should live. We could carve out another jar, but I think it makes the sense to add most of the `visitation` stuff directly to `weepickle-core`.

### DIFF
--- a/weepickle/src/main/scala/com/rallyhealth/weepickle/v1/WeePickle.scala
+++ b/weepickle/src/main/scala/com/rallyhealth/weepickle/v1/WeePickle.scala
@@ -1,12 +1,15 @@
 package com.rallyhealth.weepickle.v1
 
+import com.rallyhealth.weepickle.v1.core.ops.FromInputOpsImplicits
 import com.rallyhealth.weepickle.v1.core.{FromInput, Visitor}
 
 /**
   * Converters for default scala types.
   * Macros to generate converters for case classes.
   */
-object WeePickle extends LowPriorityImplicits {
+object WeePickle
+  extends LowPriorityImplicits
+  with FromInputOpsImplicits {
 
   object ToScala {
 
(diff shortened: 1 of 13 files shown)


weepickle/src/main/scala/com/rallyhealth/weepickle/v1/WeePickle.scala
```diff
@@ -1,12 +1,15 @@
...
 import com.rallyhealth.weepickle.v1.core.{FromInput, Visitor}
 
 /**
   * Converters for default scala types.
   * Macros to generate converters for case classes.
   */
-object WeePickle extends LowPriorityImplicits {
+object WeePickle
+  extends LowPriorityImplicits
+  with FromInputOpsImplicits {
```
htmldoug, line 12: In the future, I expect to add `FromOpsImplicits`/`ToOpsImplicits`/`VisitorOpsImplicits` for `.tolerant`, etc. Grouping the extensions into a "high level extension ops" traits seems like it should help discoverability.

arkban:
> The big decision is where it should live. We could carve out another jar, but I think it makes the sense to add most of the visitation stuff directly to weepickle-core.

+1 to put it in with the rest of the stuff. Having to manage multiple JAR versions is always a little error prone.

jducoeur (approved):

weepickle/src/main/scala/com/rallyhealth/weepickle/v1/WeePickle.scala
arkban: I think this would be a good thing to add to the code, it explains your reasoning which is often hard to "reverse engineer" from the code.

arkban (approved):

weepickle/src/main/scala/com/rallyhealth/weepickle/v1/WeePickle.scala
russellremple: also `BufferedValueOps`? https://github.com/rallyhealth/weePickle/blob/c1315d05edacb8e3dd7cb56838bc9cbfa8bfaf2f/weejson/src/main/scala/com/rallyhealth/weejson/v1/BufferedValue.scala#L35

russellremple (approved):

PR #96, opened 2021-11-12
Bugfix: MsgPackReader: properly increment index in ext

htmldoug:
Port of https://github.com/com-lihaoyi/upickle/pull/370 + new tests from upstream.


weepack/src/main/scala/com/rallyhealth/weepack/v1/MsgPackParser.scala
```diff
@@ -157,6 +157,7 @@ abstract class BaseMsgPackParser {
 
       case _ =>
```
htmldoug, line 158: Note to self: the other timestamp cases increment the index properly.

PR #98, opened 2021-12-01
perf: extract PlayJson singletons

htmldoug:
For disposable/stateful subclasses of `PlayJson`, we're wasting time recreating the `JsBoolean` singletons.

image


russellremple (approved):
lgtm

PR #99, opened 2021-12-03
Add JsValueSingletons.jsObjectEmpty

htmldoug:
Across our test dataset, 18% of all `JsObject`s are empty. 

# objects by field size in test data set

Heap sizes (🔴 and 🟢 are the PR changes):
Heap usage relative to UnorderedPlayJson (1)

Savings range from 0-15% with an average of~1%. `UnorderedPlayJson` benefits less since it already shares `EmptyMap`. Labcorp.json shaves off 248 MB out of 3,355 MB. It also cuts the DoS potential from the horrible 63x memory amplification string, `[{},{},{},...]`. Latter two examples aren't pictured in the graph.


jducoeur (approved):
Nice little optimization...

PR #101, opened 2021-12-16
Bug fixes for Play numerics

russellremple:
Added tests for Play Json translations via WeePickle From/To primitive types. Addressed various issues with handling truncation of decimals as ints or longs. Since the bugs were in `Util.scala` (in core), could address numeric handling outside of Play Json (not 100% sure, but pretty sure).

Update: added property tests for `parseIntegralNum` which exposed a few more sharp edges, e.g., where BigDecimals within the Int/Long range were not truncating properly, exploding when decimal part was too long, or not handling special cases (like `0E+50`). I think I have all the edges smoothed out now, but perhaps the whole decimal parsing logic should be revisited -- it is getting a bit unwieldy.


htmldoug (approved):
Oof!

```scala
forAll { bd: BigDecimal =>
  val n = JsNumber(bd)
  PlayJson.transform(n, PlayJson.ToJsValue) shouldBe n
}
```
Prop test seems happy with it now, whereas it failed the previous implementation.

Thanks!

htmldoug (approved):

PR #106, opened 2021-12-21
Add net/JSONTestSuite tests. Fix JsonPointer OOME on "[[[[..."

htmldoug:
Added https://github.com/nst/JSONTestSuite files and worked around `JsonPointer` requiring O(depth^2) memory. This manifested as:
- sluggish ~5s parsing on `n_structure_open_array_object.json`
- OOME on `n_structure_100000_opening_arrays.json` (exhausts `-Xmx8g`)
    Screen Shot 2021-12-21 at 2 07 32 AM


rodtan (approved):

weejson-jackson/src/main/scala/com/rallyhealth/weejson/v1/jackson/JsonParserException.scala
```diff
@@ -19,6 +22,43 @@ object JsonParserException {
...
+          } {
+            c match {
+              case '/' => sb.append("~1")
+              case '~' => sb.append("~0")
+              case _ => sb.append(c)
+            }
+          }
+        }
+        sb
+      }
```
russellremple, line 58: I love this algorithm - very efficient! I wonder, would it be any faster to skip the `.reverseIterator` and use `StringBuilder.prepend` instead of `.append`?

russellremple (approved):
lgtm, with one thing to maybe try...

weejson-jackson/src/main/scala/com/rallyhealth/weejson/v1/jackson/JsonParserException.scala
htmldoug: Thanks!

If you're talking about [insert](https://github.com/openjdk/jdk/blob/f31dead6c550444a836655ffdf97df8816e2d546/src/java.base/share/classes/java/lang/AbstractStringBuilder.java#L1090-L1100), it does a whole `System.arraycopy` to shift.

Something like an `ArrayDeque` might be interesting though.

russellremple: Yeah, I just looked at the implementation of prepended in Seq and doesn’t look efficient at all (allocates new builder on each invocation) where the implementation of reverseIterator for IndexedSeq looks great. Sure thing that taking the hit to build the ArrayBuffer once is best, so probably not worth trying to fine tune too much.

htmldoug: And the java and scala `ArrayDeque` both [box](https://github.com/scala/scala/blob/4b2151cbe4c857d49556b70c3fa0f5e236fd7754/src/library/scala/collection/mutable/ArrayDeque.scala#L39-L40), so using it for chars is a bust.

It really seems like there should be a `StringBuilderDeque` out there somewhere with O(1) prepend. Why can't I find anyone who has built one? I must not be googling hard enough....

htmldoug: Seems like we might be able to just rebuild the scala one with `@specialized`.

htmldoug: A challenge with a hypothetical `StringBuilderArrayDeque` would be getting the end result into a `String`. We could efficiently copy the end result to a `char[]` with two `System.arraycopy`, but there's no way to convince `new String(char[])` avoid making another copy of it. So we end up with 3 copies (ArrayDeque, char[], String), whereas the `StringBuilder` can do it in two.

We could try to reorder the Deque's array in place, but we'd need buffer space if we want to do it fast with `System.arraycopy`. Could possibly use the extra capacity if available. Otherwise, we need an extra buffer (maybe a `ThreadLocal` pooled one would only be viable for small ones). Last simple option is to swap element by element, but that might be slower than arraycopy.

Interesting rabbit hole.

PR #109, opened 2022-01-05
FromJson(String, etc.).transform() replayable

### LINKED ISSUE #108: Throw a more meaningful message when transforming FromInput more than once
`FromInput` can be transformed only once:
```scala
val fromInput: FromInput = FromJson("""{"a":"1"}""")
fromInput.transform(ToScala[Test])
fromInput.transform(ToJson.string)
fromInput.transform(ToValue)
fromInput.transform(ToJsValue)
```

second transformation (order of Visitors doesn't matter) always fails with the following error:

```
// weepickle 1.6.0
com.fasterxml.jackson.core.io.JsonEOFException: There were no tokens to parse
 at [Source: (String)"{"a":"1"}"; line: 1, column: 10]
  at com.rallyhealth.weejson.v1.jackson.JsonFromInput.transform(FromJson.scala:55)
  ... 36 elided

// weepickle 1.7.1
com.rallyhealth.weepickle.v1.core.TransformException: Premature EOF jsonPointer= index=9 line=1 col=10 token=null
  at com.rallyhealth.weejson.v1.jackson.JsonParserException$.apply(JsonParserException.scala:14)
  at com.rallyhealth.weejson.v1.jackson.JsonFromInput.parseRec(FromJson.scala:95)
  at com.rallyhealth.weejson.v1.jackson.JsonFromInput.transform(FromJson.scala:59)
  ... 36 elided

```

Please throw a more meaningful message when transforming `FromInput` more than once (e.g. `Input is already transformed`).

Transforming of value many times is ok:
```scala
val jsonValue: Value = FromJson("""{"a":"1"}""").transform(ToValue)
jsonValue.transform(ToJson.string)
jsonValue.transform(ToScala[Test])
jsonValue.transform(ToValue)
jsonValue.transform(ToJsValue)
```

htmldoug:
Allow calling `.transform()` on a `FromJson`, `FromYaml`, `FromXml`, etc. that was created with a replayable input type (`String`, `Array[Byte]`, `File`, etc.). Otherwise, throws a clearer exception if the `JsonParser` has already been closed.

Fixes #108. @marekklis

### DIFF
--- a/weejson-jackson/src/main/scala/com/rallyhealth/weejson/v1/jackson/FromJson.scala
+++ b/weejson-jackson/src/main/scala/com/rallyhealth/weejson/v1/jackson/FromJson.scala
@@ -4,7 +4,7 @@ import com.fasterxml.jackson.core.JsonParser.NumberType
 import com.fasterxml.jackson.core.JsonToken._
 import com.fasterxml.jackson.core.JsonTokenId._
 import com.fasterxml.jackson.core.{JsonFactory, JsonParser, JsonToken, JsonTokenId}
-import com.rallyhealth.weepickle.v1.core.{CallbackVisitor, FromInput, TransformException, Visitor}
+import com.rallyhealth.weepickle.v1.core.{Abort, CallbackVisitor, FromInput, TransformException, Visitor}
 
 import java.io.{File, InputStream, Reader}
 import java.nio.file.Path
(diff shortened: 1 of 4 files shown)


weejson-jackson/src/main/scala/com/rallyhealth/weejson/v1/jackson/FromJson.scala
```diff
@@ -29,32 +29,38 @@ object FromJson extends JsonParserOps {
...
 
   def apply(in: InputStream): FromInput = fromParser(factory.createParser(in))
 
   def apply(reader: Reader): FromInput = fromParser(factory.createParser(reader))
 
-  def apply(file: File): FromInput = fromParser(factory.createParser(file))
+  def apply(file: File): FromInput = fromReplayableParser(() => factory.createParser(file))
 
-  def apply(path: Path): FromInput = fromParser(factory.createParser(path.toFile))
+  def apply(path: Path): FromInput = fromReplayableParser(() => factory.createParser(path.toFile))
```
danny-goldberg-rally, line 42: would a temp file, i.e. an intermediate result from some computation that then gets cleaned up, be an edge case to be concerned about?

in that case, would it simply hit `if (parser.isClosed) throw new Abort("Parser is closed.")`?

danny-goldberg-rally (approved):

weejson-jackson/src/main/scala/com/rallyhealth/weejson/v1/jackson/FromJson.scala
htmldoug: Good question. Parser probably wouldn't be closed--I think the only way to do that is to call `parser.close()` explicitly. I'd expect a jackson EOF or maybe `IOException` in that case. Regardless, I'm not sure there's anything we should do about it.

marekklis (approved):

PR #112, opened 2022-05-17
Fix None macro scoping

devansc:
An attempt to solve this bug:
image

worksheet code:
```
import com.rallyhealth.weejson.v1.jackson.FromJson

object BreakMe  {
  val None = 1
}
import BreakMe._

case class Foo(
  x: Int,
  y: Option[String]
)

object Foo {
  implicit val fromTo = macroFromTo[Foo]
}


FromJson("""{"x": 123}""").transform(ToScala[Foo])
```

### DIFF
--- a/weepickle-implicits/src/main/scala-2/com/rallyhealth/weepickle/v1/implicits/MacroImplicits.scala
+++ b/weepickle-implicits/src/main/scala-2/com/rallyhealth/weepickle/v1/implicits/MacroImplicits.scala
@@ -168,7 +168,7 @@ private object MacroImplicits {
         if (isParamWithDefault) {
           Some(q"$companion.$defaultName")
         } else {
-          if (assumeDefaultNone) Some(q"${TermName("scala.None")}")
+          if (assumeDefaultNone) Some(q"scala.None")
           else None
         }
       }
--- a/weepickle-implicits/src/main/scala-2/com/rallyhealth/weepickle/v1/implicits/MacroImplicits.scala
+++ b/weepickle-implicits/src/main/scala-2/com/rallyhealth/weepickle/v1/implicits/MacroImplicits.scala
@@ -168,7 +168,7 @@ private object MacroImplicits {
         if (isParamWithDefault) {
           Some(q"$companion.$defaultName")
         } else {
-          if (assumeDefaultNone) Some(q"${TermName("None")}")
+          if (assumeDefaultNone) Some(q"${TermName("scala.None")}")
           else None
         }
       }
--- a/weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/macros.scala
+++ b/weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/macros.scala
@@ -56,7 +56,7 @@ def getDefaultParamsImpl[T](using Quotes, Type[T]): Expr[String => Option[() => 
     val namesExpr: Expr[List[String]] = Expr.ofList((names ++ assumeDefaultNoneNames).map(Expr(_)))
     val identsExpr: Expr[List[() => AnyRef]] = Expr.ofList(
       idents.map(_.asExpr).map(i => '{ () => $i.asInstanceOf[AnyRef] }) ++
-      assumeDefaultNoneNames.map(_ => '{ () => None })
+      assumeDefaultNoneNames.map(_ => '{ () => scala.None })
     )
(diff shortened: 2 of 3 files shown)


htmldoug:
Thanks! Is this fine in scala 3? Can we get a test that covers both of them?

devansc:
> Thanks! Is this fine in scala 3? Can we get a test that covers both of them?

Unfortunately this change doesn't fix it... I was trying to get a build going to test it in a downstream service, but yeah I realize now unit tests break with this change. 

Not super familiar with this code and macros in general. So mainly put this up to make awareness of the issue and see if anyone had any suggestions. I might try to keep poking around a bit, but I at least unblocked myself with a downstream fix. I feel this should be addressed still though. Potentially I can just convert this to an Issue if I don't get around to figuring it out.

Maybe @russellremple would have an idea though...

russellremple:
Weird, the change seems right to me and I would have expected this to pass -- not sure what is going on. Location requiring a similar change in Scala 3 macros is in https://github.com/rallyhealth/weePickle/blob/v1/weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/macros.scala#L59. However, there are a few places where macros check stringified representations of types, so I think you could always create some kind of pathological case that would confuse them. (E.g., create your own type called "Option" would probably also break things.)

russellremple:
I believe the solution is, if `None` is fully qualified, it doesn't have to be a `TermName` anymore, and it can just be parsed as code. So you should be able to just say `if (assumeDefaultNone) Some(q"scala.None")`. Similarly, in the Scala 3 case, you should be able to say `assumeDefaultNoneNames.map(_ => '{ () => scala.None })`.

weepickle-implicits/src/main/scala-2/com/rallyhealth/weepickle/v1/implicits/MacroImplicits.scala
```diff
@@ -168,7 +168,7 @@ private object MacroImplicits {
         if (isParamWithDefault) {
           Some(q"$companion.$defaultName")
         } else {
-          if (assumeDefaultNone) Some(q"${TermName("None")}")
+          if (assumeDefaultNone) Some(q"${TermName("scala.None")}")
```
russellremple: Try `if (assumeDefaultNone) Some(q"scala.None")` instead.

devansc:
The test properly fails with the old code:
Screen Shot 2022-05-17 at 2 40 07 PM
And passes with new code.

htmldoug (approved):

weepickle-implicits/src/main/scala-3/com/rallyhealth/weepickle/v1/implicits/macros.scala
```diff
@@ -56,7 +56,7 @@ def getDefaultParamsImpl[T](using Quotes, Type[T]): Expr[String => Option[() =>
     val namesExpr: Expr[List[String]] = Expr.ofList((names ++ assumeDefaultNoneNames).map(Expr(_)))
     val identsExpr: Expr[List[() => AnyRef]] = Expr.ofList(
       idents.map(_.asExpr).map(i => '{ () => $i.asInstanceOf[AnyRef] }) ++
-      assumeDefaultNoneNames.map(_ => '{ () => None })
+      assumeDefaultNoneNames.map(_ => '{ () => scala.None })
```
htmldoug: is this one actually necessary? Seems like `None` should be unambiguous in this context.

devansc: I will resort to @russellremple on this one... I don't actually know if there are scala3 tests and if I can add something to test if it's necessary. Either way, it certainly doesn't hurt...

htmldoug: The `ExampleTests` addition you made covers both scala 2 and 3. If removed, does the test pass?

russellremple: You are right, it isn't needed for Scala 3. Can you explain why "`None` should be unambiguous in this context"? You are 100% correct, I just don't understand.

russellremple (approved):
lgtm

htmldoug (approved):

russellremple (approved):

PR #113, opened 2022-06-28
Add From/ToXml to readme

### LINKED ISSUE #111: Make XML support more discoverable
After reading this project readme I supposed it is ready to use to parse XML encoded data:

> A stable JSON, YAML, MsgPack, XML, etc. serialization framework based

I did not find any documentation on how to use XML, and v1.jackson package does not seem to contain any XML code. If there is no real XML support, please, do not advertise it in the readme.

See also https://users.scala-lang.org/t/question-on-xml-to-json-with-weepickle-or-upickle/6512/2:

> The XML aspect is a little aspirational, I’m afraid – it should be straightforward in theory, but nobody’s actually done it yet.

htmldoug:
@OndrejSpanel, how's this look? Can you think of any other info that should be included?

Fixes #111.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -33,17 +33,17 @@ libraryDependencies += "com.rallyhealth" %% "weepickle-v1" % "version"
 ```
 
 ## Getting Started
-JSON to Scala:
+#### JSON to Scala
 ```scala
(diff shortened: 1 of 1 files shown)


OndrejSpanel:
Perfect, thanks.

PR #114, opened 2022-06-28
Use properly decoded field names when handling case classes

### LINKED ISSUE #110: hyphens in identifier name don't work with serialization/de-serialization
I have a case class here with a hyphen in the identifier:
```
case class test(
  `name-withhyphen`: String
)

object test {
  implicit val rw: FromTo[test] = macroFromTo
}
```
When serializing/deserializing like this, I get:

```
val request = {"name-withhyphen":"name"}
request.validate(ToScala[test])
```

I get this error: `+com.rallyhealth.weepickle.v1.core.Abort: missing keys in dictionary: name$minuswithhyphen []`

htmldoug:
Port of https://github.com/com-lihaoyi/upickle/pull/366. Fixes #110.


russellremple (approved):
lgtm

PR #116, opened 2022-08-23
Add to/from java.time.ZoneId

compmstr:
I came across a use case where I needed to (de)serialize `java.time.ZoneId`s.  

Since a lot of other java.time classes were already supported, I went ahead and added ZoneId.

### DIFF
--- a/weepickle-implicits/src/main/scala/com/rallyhealth/weepickle/v1/implicits/Froms.scala
+++ b/weepickle-implicits/src/main/scala/com/rallyhealth/weepickle/v1/implicits/Froms.scala
@@ -15,4 +15,5 @@ trait Froms extends DefaultFroms { this: Annotator =>
     override def transform0[V](v: Instant, out: Visitor[_, V]): V = out.visitTimestamp(v)
   }
   implicit val FromDate: From[Date] = FromInstant.comap(_.toInstant)
+  implicit val FromZoneId: From[ZoneId] = FromString.comap[ZoneId](_.toString)
 }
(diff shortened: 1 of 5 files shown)


weepickle-implicits/src/main/scala/com/rallyhealth/weepickle/v1/implicits/Froms.scala
```diff
@@ -15,4 +15,5 @@ trait Froms extends DefaultFroms { this: Annotator =>
     override def transform0[V](v: Instant, out: Visitor[_, V]): V = out.visitTimestamp(v)
   }
   implicit val FromDate: From[Date] = FromInstant.comap(_.toInstant)
+  implicit val FromZoneId: From[ZoneId] = FromString.comap[ZoneId](_.toString)
```
htmldoug: I don't think adding members to traits will pass MiMa, but they can go into `LowPriorityImplicits` instead since it's a class. The end result should be the same.

compmstr: Moved the new implicits into `LowPriorityImplicits`

htmldoug (approved):
thanks!

weepickle-tests/src/test/scala/com/rallyhealth/weepickle/v1/example/JvmPrimitiveTests.scala
```diff
@@ -30,6 +30,11 @@ object JvmPrimitiveTests extends TestSuite {
       test("Date") {
         rw(new Date(1579664660824L), "\"2020-01-22T03:44:20.824Z\"")
       }
+      test("ZoneId") {
+        rw(ZoneId.of("America/New_York"), "\"America/New_York\"")
+        rw(ZoneId.of("US/Eastern"), "\"US/Eastern\"")
+        rw(ZoneId.of("US/Pacific"), "\"US/Pacific\"")
```
htmldoug, line 36:
```suggestion
        rw(ZoneId.of("America/New_York"), "\"America/New_York\""))
        rw(ZoneId.of("US/Eastern"), "\"US/Eastern\""))
        rw(ZoneId.of("US/Pacific"), "\"US/Pacific\""))
```

russellremple (approved):
lgtm

PR #121, opened 2023-05-06
Scala 3.3/Play 2.9 compatibility

russellremple:
## Proposed changes

Make sure we are using play-json versions compatible with Play 2.9, and include an artifact for Scala 2.13. Upgrade upickle and sourcecode. Update Scala 3 to LTS version.

Play 2.9 will use play-json 2.10, which is, after a very long release candidate phase, is now available. This PR updates the Scala (LTS) and play-json versions. (Note that Play 2.9 is still a milestone release, but should be out soon.) It also addresses some deprecations and updates various libraries/plugins to the latest.
## Types of changes

What types of changes does your code introduce?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the [CONTRIBUTING](https://github.com/rallyhealth/weePickle/blob/v1/CONTRIBUTING.md) doc 
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in downstream modules

## Further comments

Play 2.9 will use play-json 2.10, which is, after a very long release candidate phase, now generally available. This PR updates Scala (LTS) and play-json versions. (Note that Play 2.9 is still a milestone release, but should be out soon.)


weejson-jackson/src/main/scala/com/rallyhealth/weejson/v1/jackson/VisitorJsonGenerator.scala
```diff
@@ -28,7 +28,7 @@ class VisitorJsonGenerator[J](
 
   private val cs = new TextBufferCharSequence(Array.emptyCharArray, 0, 0)
 
-  private val stack: mutable.ArrayStack[ObjArrVisitor[Any, _]] = new mutable.ArrayStack[ObjArrVisitor[Any, _]]()
+  private var stack: List[ObjArrVisitor[Any, _]] = Nil
```
htmldoug: This will slightly increase memory allocation rate (for all but the smallest messages), although probably acceptable.

weejson/src/main/scala-3/com/rallyhealth/weejson/v1/package.scala
```diff
@@ -0,0 +1,6 @@
+package com.rallyhealth.weejson
+
+package object v1 {
+  private val CollectionConverters = scala.jdk.CollectionConverters
```
htmldoug, line 4: Maybe a dumb question, but will the scala 2/3 compilers both be okay with namespace collisions if weepickle users `import com.rallyhealth.weejson.v1._`? I know it's a `private val`, but I've seen some strange compiler bugs.

A more paranoid alternative might be to tuck this away in an `weejson.v1.internal` or `weejson.v1.compat` package.

Please disregard if you think it's a non-issue or more trouble than it's worth.

htmldoug (approved):

weejson-jackson/src/main/scala/com/rallyhealth/weejson/v1/jackson/VisitorJsonGenerator.scala
russellremple: @htmldoug, so good to hear from you (I did not expect it)!!!

I was pretty much blindly following the direction of the compiler(s) here - if you think the change will be for the worse, I will revert it and ignore the deprecation. Scala 2.13 (and I assume Scala 3 as well) deprecated `ArrayStack` with the message "Use `Stack` instead of `ArrayStack`; it now uses an array-based implementation". But if I do that, then Scala 2.12 complains that "`Stack` in package mutable is deprecated (since 2.12.0): `Stack` is an inelegant and potentially poorly-performing wrapper around `List`. Use a `List` assigned to a var instead." So that's what I did.

weejson/src/main/scala-3/com/rallyhealth/weejson/v1/package.scala
russellremple: A few months back I integrated a locally-published version of this PR with quite a bit of client code, both Scala 2 and 3, and did not see any collisions. I've validated that a lot of that code has wildcard imports exactly as you describe. So I don't believe this will be an issue.

weejson-jackson/src/main/scala/com/rallyhealth/weejson/v1/jackson/VisitorJsonGenerator.scala
htmldoug: Good to hear from you, too!

If you want to be diligent, you could measure before/after. Since scala reimplemented `Stack` to use `Array` instead of `List`, maybe performance was one of the reasons.

IIRC, the difference in throughput was negligible in benchmarks, although maybe there are different conditions where it matters. I kinda went after all allocation indiscriminately since it always eliminated GC pauses, sometimes improved throughput, and most of all, removed allocation noise from the profiler that I no longer had to scrutinize further as "okay" or "bad".

Worst case scenario, it's `private`, so it can be changed later without breaking binary compatibility.

I'd say go ahead using the `List`. Admittedly, I have much less skin in the game now that I'm no longer a service owner.

build.sbt
```diff
@@ -9,20 +9,20 @@ crossScalaVersions := Nil // crossScalaVersions must be set to Nil on the aggreg
...
     `weepickle-tests` % "compile;test",
-    `weejson-upickle`,
   )
   .enablePlugins(JmhPlugin)
   .settings(
     noPublish,
     crossScalaVersions := Seq(scala213, scala3),
     libraryDependencies ++= Seq(
+      "com.lihaoyi" %% "upickle" % "3.1.0", // need the latest Scala 3 support
       "com.fasterxml.jackson.core" % "jackson-databind" % "2.13.0",
```
perryjp: Should we bump the version of this (and any others?) 2.13.0 has 4 known vulnerabilities

htmldoug: Sure.

russellremple: Take a look at https://github.com/rallyhealth/weePickle/pull/121/commits/b83a3f346576487f7f46cc9464f95b81049a95c5. I did what I could (and still stay binary compatible).

build.sbt
```diff
@@ -9,27 +9,27 @@ crossScalaVersions := Nil // crossScalaVersions must be set to Nil on the aggreg
...
     libraryDependencies ++= Seq(
-      "com.fasterxml.jackson.core" % "jackson-databind" % "2.13.0",
-      "com.fasterxml.jackson.dataformat" % "jackson-dataformat-smile" % "2.13.0",
-      "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.13.0",
-      "io.circe" %% "circe-generic" % "0.14.1",
-      "io.circe" %% "circe-parser" % "0.14.1",
-      "org.msgpack" % "jackson-dataformat-msgpack" % "0.9.0",
-      "com.lihaoyi" %% "sourcecode" % "0.2.7",
+      "com.lihaoyi" %% "upickle" % "3.1.3", // need the latest Scala 3 support
+      "com.fasterxml.jackson.core" % "jackson-databind" % "2.15.2",
```
perryjp, line 19: I actually wonder if this is updating too far, I notice play-json is only on version 2.14.3. Any idea how closely we need to track them @htmldoug ?

perryjp: (I know databind can get picky about versions, but maybe the “binary compatible” check makes sure this doesn’t go too far?)

russellremple: That is a really good point, and ratcheting back to 2.14.3 might be a good idea, but ratcheting all the way back to 2.11.4 (from `play-json` 2.8.2) or 2.9.8 (from `play-json` 2.7.4) is probably not. Thankfully the direct dependencies on Jackson's `-databind`, `-dataformat-smile`, `-dataformat-msgpack`, and `-module-scala` are limited to the unpublished `bench` project. The only direct Jackson dependencies from published projects are on `-core` (by `weejson`, `weexml`, and `weeyaml`), `-dataformat-xml` (by `weexml`), and `-dataformat-yaml` (by `weeyaml`). However, both `-dataformat-xml` and `-dataformat-yaml` pull in `-databind` transitively, but I don't believe `weexml` and `weeyaml` are widely used, and if someone needs to depend on the older version, they can exclude/override the transitive one.

perryjp: I've been looking around and I see some places that use weePickle that have updated databind to 2.15.2 so I think it's probably not an issue.

perryjp (approved):

russellremple:
@htmldoug I really need your help. I just realized that I have _no idea_ how to release 1.9.0 for this (and I need to). It isn't documented in our README and there's [this](https://github.com/rallyhealth/weePickle/blob/v1/project/SonatypePlugin.scala#L32) which makes me think some of it may still be tied to you personally. Could you please "teach me how to fish" in this context?

htmldoug:
I used sbt-release for workflow. I looked up instructions on their readme every time.

You'll have to get maven central credentials. IIRC, I had to create a jira ticket.

